PR #2, opened 2021-03-15
[38398687]: Add UI for creating new personal PastelId and registring it as ID ticket

br4in3x:
https://freedcamp.com/view/2956183/tasks/38398687

Adds UI for the Pastel ID UI functionality.

This will allow creating PastelIDs via RPC. However, the functionality to actually create a ticket and charge for the creation of the PastelID will be implemented in https://freedcamp.com/view/2956183/tasks/38401246.

### DIFF
--- a/app/Routes.js
+++ b/app/Routes.js
@@ -33,6 +33,7 @@ import AddressbookImpl from './utils/AddressbookImpl';
 import Sidebar from './components/Sidebar';
 import Transactions from './components/Transactions';
 import CompanionAppListener from './companion';
+import PastelID from './components/PastelID';
 import WormholeConnection from './components/WormholeConnection';
 
 type Props = {};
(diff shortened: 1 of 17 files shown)


app/Routes.js
```diff
@@ -33,6 +33,7 @@ import AddressbookImpl from './utils/AddressbookImpl';
 import Sidebar from './components/Sidebar';
 import Transactions from './components/Transactions';
 import CompanionAppListener from './companion';
+import PastelID from './components/PastelID';
```
Dicklesworthstone, line 36: I think we should change this to be Pastel_ID instead because it looks too similar to Pasteld

Dicklesworthstone (approved):

Dicklesworthstone:
Looks fine to me.

app/Routes.js
br4in3x: Addressed according to conversation in Slack.

andriy-shtanheiev (approved):

PR #3, opened 2021-03-29
[38737877]: Fix build Pastel process

br4in3x:
https://freedcamp.com/dashboard/home/panel/task/38737877

Fixes pastel build process by leveraging electron-packager instead of electron-builder.

Electron-packager comes in a bundle with electron-forge, so it's always good to use tools that are meant to work together.


br4in3x:
Merging this PR without any reviews because these changes were supposed to be a part of the new main branch `master`.

PR #6, opened 2021-03-31
[38737882]: Switch wallet codebase to use Redux for state management

br4in3x:
https://freedcamp.com/view/2956183/tasks/panel/task/38737882

Adds Redux and Redux Toolkit to the wallet, and sets up project structure based on "features".

This PR will enable app-wide state management with Redux and Redux Toolkit. This is a very well understood design pattern that will reduce complexity in the app.


package.json
```diff
@@ -58,6 +26,11 @@
     "@electron-forge/maker-squirrel": "6.0.0-beta.54",
     "@electron-forge/maker-zip": "6.0.0-beta.54",
     "@electron-forge/plugin-webpack": "6.0.0-beta.54",
+    "@types/electron-devtools-installer": "^2.2.0",
```
namnm, line 29: Can we remove the ^ in version number, so that we would have exact version installed. This can help prevent some unexpected minor version changed, which cause some issue.

After remove all the ^ you may need to run `yarn` to update `yarn.lock` as well

namnm (changes requested):

namnm (approved):

PR #7, opened 2021-03-31
Convert old js files into ts

namnm:


src/components/LoadingScreen.tsx
```diff
@@ -139,15 +132,18 @@ class LoadingScreen extends Component {
...
 
-    file.on('finish', () => file.close(cb)) // check for request errors
+    file.on('finish', () => {
+      file.close()
+      cb()
+    }) // check for request errors
 
     sendReq.on('error', err => {
-      fs.unlink(dest)
+      fs.promises.unlink(dest)
```
br4in3x, line 141: I think changing this is not safe. I woudn't change `fs.unlink` to `fs.promises.unlink`.

src/components/LoadingScreen.tsx
```diff
@@ -139,15 +132,18 @@ class LoadingScreen extends Component {
...
       return cb(err.message)
     })
     file.on('error', err => {
       // Handle errors
-      fs.unlink(dest) // Delete the file async. (But we don't check the result)
+      fs.promises.unlink(dest) // Delete the file async. (But we don't check the result)
```
br4in3x, line 146: Same as above

src/components/Transactions.tsx
```diff
@@ -16,6 +11,7 @@ import ScrollPane from './ScrollPane'
 import Utils from '../utils/utils'
 import AddressBook from './Addressbook'
 import routes from '../constants/routes.json'
+import { PastelURITarget } from '../utils/uris'
```
br4in3x, line 14: This is also not safe. I understand that this import was missing, but fixing bugs in legacy code can be punishing. Is it possible not to import this and not trigger a ts error?

br4in3x (dismissed):

PR #9, opened 2021-04-01
Update CONTRIBUTING.md

namnm:


docs/CONTRIBUTING.MD
```diff
@@ -1,11 +1,19 @@
...
-To start for development, run:
+Pastel Wallet Fullnode is an Electron app built with React on the frontend.
 
-```
-yarn start
-```
+##### Getting started
 
-### Important
+You should have Node LTS and Yarn installed globally. After project clone, run `yarn` to install the dependencies. To run the daemon service `pasteld`, you can select the proper binary in `static/bin`, or just download and run the existing official Pastel Wallet Fullnode app from website. Finally run `yarn start` to start the Electron app locally in development mode.
```
br4in3x, line 7: You don't need to start `pasteld` manually, the app will launch it. Here is where it happens: https://github.com/pastelnetwork/pastel-electron-wallet/blob/master/src/components/LoadingScreen.tsx#L341

docs/CONTRIBUTING.MD
```diff
@@ -1,11 +1,19 @@
...
 
-Please note that this is a fork of another wallet, that has been written using Flow and moved over to TypeScript. During development, under no condition we should create any javascript files, and use TypeScript only from this stage.
+##### Code linting and standard checks
+
+All the code in this project are written in TypeScript with some linting and standard checks. New features and code must follow this same standard. We recommend using Visual Studio Code IDE for working on this code base. Some available commands can be run:
+
+- `yarn format` to automatically format all code in the project.
+- `yarn check-all` to perform all the checks, mostly used in the CI.
+- `yarn check-type-coverage` to check and show if all the code are strong typed or not. Since we have been migrating from the old code, this check hasn't passed yet and not in the above `check-all` command. We can work on improving those old code part-by-part by using this command, it will show us all the issues and where we need to fix.
```
br4in3x, line 15: Could you make them all separate commands and not including `check-all` command? The reason for that is, it will allow to keep us in check on what is running on the CI by explicitly saying what we need, and also it would keep the number of scripts low, because currently there are too many of them.

br4in3x (commented):
Nice document!

docs/CONTRIBUTING.MD
Dicklesworthstone: This links gives me a 404

Dicklesworthstone (approved):

br4in3x (approved):

PR #10, opened 2021-04-02
chore: Move old files to a legacy folder

br4in3x:
Moves old files into `src/legacy`.

This will discourage new developers from changing or adding code in the old files that do not follow our coding standards.


Dicklesworthstone:
Sounds like a good idea.

Dicklesworthstone (approved):

namnm (approved):

PR #15, opened 2021-04-08
Add paper wallet generator

ngvtuan:
##### Commits
- [x] Add paper wallet generator. Note: This task is not fully completed the PR is only for code review purpose.
- [x] Remove console and Let's have an exact version
- [x] Update the code follow PR’s comment
- [x] Removed right click
- [x] Add end line to pdf

##### FreedCamp
- [x] [Add Paper Wallet Generator] (https://freedcamp.com/view/2956183/tasks/38537599)


src/features/pastelPaperWalletGenerator/pastelPaperWalletGenerator.tsx
```diff
@@ -0,0 +1,177 @@
...
+    return (
+      <Svg
+        shape-rendering='crispEdges'
+        height='100%'
+        width='100%'
+        viewBox='0 0 33 33'
+      >
+        {paths &&
+          paths.length > 0 &&
+          paths?.map((path, idx) => (
```
br4in3x, line 117: Since paths are present (line 115), optional changing is not required here

package.json
```diff
@@ -28,6 +28,7 @@
   },
   "dependencies": {
     "@fortawesome/fontawesome-free": "5.12.0",
+    "@react-pdf/renderer": "^2.0.0",
```
br4in3x, line 31: Let's have an exact version here please

src/legacy/components/Receive.tsx
```diff
@@ -59,7 +62,9 @@ const AddressBlock = ({
     >
       <AccordionItemHeading>
         <AccordionItemButton className={cstyles.accordionHeader}>
-          {address}
+          <span onClick={() => handleShowPastelPaperWalletModal(address)}>
+            {address}
+          </span>
```
br4in3x, line 67: Not sure if clicking on the wallet address is a good idea, because this action is already taken: when you click an address, an accordion item expands. Perhaps adding another button next to "Copy address" and "Export Viewing Key" is a good idea?

src/features/pastelPaperWalletGenerator/pastelPaperWalletGenerator.tsx
```diff
@@ -0,0 +1,177 @@
...
+    )
+  }
+
+  return (
+    <Modal
+      isOpen={props.modalIsOpen}
+      onRequestClose={() => props.onCloseModal()}
+    >
+      <div className={styles.modal_content}>
+        <PDFViewer width='100%' height='100%'>
```
br4in3x, line 134: Would it be possible to hide a preview on the left section of the PDF viewer? It takes a lot of space and has no value

br4in3x, line 134: Would it be possible to move this width and height parameters to the css module?

src/features/pastelPaperWalletGenerator/pastelPaperWalletGenerator.tsx
```diff
@@ -0,0 +1,177 @@
...
+    },
+    topImg: {
+      width: '100%',
+    },
+    mainImg: {
+      width: '100%',
+    },
+  })
+
+  function generateQrcode() {
```
br4in3x, line 101: I think we should come up with a way to safely fail if for some reason `ReactDOMServer.renderToString` or `<QRCode value={props.address} renderAs='svg' />` fails. Maybe a `try/catch` or an error boundaries component?

src/features/pastelPaperWalletGenerator/pastelPaperWalletGenerator.tsx
```diff
@@ -0,0 +1,177 @@
...
+              d={path.replace('d="', '')}
+              key={idx}
+            ></Path>
+          ))}
+      </Svg>
+    )
+  }
+
+  return (
+    <Modal
```
br4in3x, line 129: When the modal opens, it expands for almost to the full width, and it's a bit difficult to close. Perhaps we can provide some closing mechanism?

namnm (changes requested):
Need to pass CI lint and test

src/features/pastelPaperWalletGenerator/pastelPaperWalletGenerator.tsx
```diff
@@ -0,0 +1,177 @@
...
+  address: string
+  privateKey: string
+  modalIsOpen: boolean
+  onCloseModal: () => void
+}
+
+function PastelPaperWalletGenerator(
+  props: PastelPaperWalletGeneratorProps,
+): JSX.Element {
+  console.log(props)
```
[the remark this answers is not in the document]
ngvtuan: I have removed it @br4in3x

src/features/pastelPaperWalletGenerator/pastelPaperWalletGenerator.tsx
ngvtuan: it has been updated

package.json
ngvtuan: Updated

src/legacy/components/Receive.tsx
ngvtuan: I have added the right click context menu for this one

src/features/pastelPaperWalletGenerator/pastelPaperWalletGenerator.tsx
ngvtuan: I'm researching on this one

ngvtuan: it has been updated

src/features/pastelPaperWalletGenerator/pastelPaperWalletGenerator.tsx
ngvtuan: I have added try/catch

src/features/pastelPaperWalletGenerator/pastelPaperWalletGenerator.tsx
ngvtuan: Added close button

src/features/pastelPaperWalletGenerator/pastelPaperWalletGenerator.tsx
```diff
@@ -0,0 +1,194 @@
...
+      const fills = qrcodeValue.match(/fill="([^"]*)/g)
+
+      return (
+        <Svg
+          shape-rendering='crispEdges'
+          height='100%'
+          width='100%'
+          viewBox='0 0 33 33'
+        >
+          {paths &&
```
namnm, line 121: This can be changed to `paths?.map(...)`

src/legacy/components/Receive.tsx
```diff
@@ -333,6 +347,15 @@ export default class Receive extends Component<any> {
             </TabPanel>
           </Tabs>
         </div>
+        <PastelPaperWalletGenerator
+          address={selectedAddress}
+          privateKey={
+            selectedAddressPrivateKeys &&
```
namnm, line 353: This can be changed to `selectedAddressPrivateKeys?.[selectedAddress]`

namnm (changes requested):

namnm (approved):

src/legacy/components/Receive.tsx
```diff
@@ -203,6 +199,16 @@ const AddressBlock = ({
...
               )}
+
+              <button
+                className={[cstyles.primarybutton, styles.buttomMarginTop].join(
+                  ' ',
+                )}
+                type='button'
+                onClick={() => fetchAndSetSinglePrivKey(address, 'paperWallet')}
+              >
+                Turn address into paper wallet
```
br4in3x, line 210: Can we change it to "Generate paper wallet" please?

br4in3x (changes requested):
Could we please move all state and state related logic into a slice, like it's done for the errorModal, for example? We don't want to keep the legacy pattern of passing state mutation functions down to child components. In the end, we want to see all the new code for paperWalletGenerator in the `features/pastelPaperWalletGenerator` folder, and legacy code modified _only_ if it's necesarry, for example, to mount the `PastelPaperWalletGenerator` component.

src/features/pastelPaperWalletGenerator/pastelPaperWalletGenerator.tsx
```diff
@@ -0,0 +1,192 @@
...
+    },
+    topImg: {
+      width: '100%',
+    },
+    mainImg: {
+      width: '100%',
+    },
+  })
+
+  function generateQrcode() {
```
br4in3x, line 106: Let's separate this into its own component, and call it something like `QRCodeGEnerator`

src/legacy/Routes.tsx
```diff
@@ -57,6 +57,9 @@ class RouteApp extends React.Component<any, any> {
       errorModalData: new ErrorModalData(),
```
br4in3x, line 57: All the state changes in this file are the legacy antipattern we want to avoid. Let's have all this moved into a redux reducer ("slice" in terms of @reduxjs/toolkit), and here, if necessary, call `dispatch` functions to populate QRCodeGenerator state.

src/features/pastelPaperWalletGenerator/pastelPaperWalletGenerator.tsx
```diff
@@ -0,0 +1,192 @@
...
+import styles from './PastelPaperWalletGenerator.module.css'
+
+type PastelPaperWalletGeneratorProps = {
+  address: string
+  privateKey: string
+  modalIsOpen: boolean
+  onCloseModal: () => void
+}
+
+function PastelPaperWalletGenerator(
```
br4in3x, line 25: Strictly speaking, this is a `PastelPaperWalletModal`, and the `generateQrcode` function is actually a `QRCodeGenerator`. You can change the component names as you like, I just wanted to give an idea how they should look like

br4in3x (changes requested):

src/legacy/components/Receive.tsx
ngvtuan: @br4in3x it has been updated

src/features/pastelPaperWalletGenerator/pastelPaperWalletGenerator.tsx
ngvtuan: It has been updated @br4in3x

src/legacy/Routes.tsx
ngvtuan: @br4in3x i have updated follow your mention

src/features/pastelPaperWalletGenerator/pastelPaperWalletGenerator.tsx
ngvtuan: @br4in3x yeah i have changed the component names as your suggest

br4in3x (approved):

PR #16, opened 2021-04-09
Release 1.1.1

br4in3x:
This will create a new 1.1.1 release.

The "Export All Private Keys" and "Pastel ID" sections are hidden from the wallet because:
1. Export All Private Keys is not currently working
2. Pastel ID is not currently completed


Dicklesworthstone (approved):

namnm:
Release should not be merged into master

PR #20, opened 2021-04-11
Copy private key instead of showing it

namnm:
Copy private key instead of showing it

Fix: https://freedcamp.com/view/2956183/tasks/panel/task/39169913


Dicklesworthstone (dismissed):

src/legacy/components/Receive.tsx
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable */
```
br4in3x, line 1: Since there are so many changes in this legacy file, would you please move it to the `features` folders, so it's not legacy anymore? There is already a lot of work done on it, and it's not really a legacy anymore, except it is still a `class` component and its state and props are untyped.

namnm: Those tasks are in urgent required by Jeff so that why I worked on weekend so we can include those fixes in the new release. You can add new tasks separately about moving and refactoring in Freedcamp. Let's just review and test this to move forward for the release.

Dicklesworthstone (approved):

src/features/recieve/AddressBlock.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+  // deprecated, see Utils.getPslToUsdString
+  pslPrice: number
+}
+export interface IAddressBlockState {
+  copiedAddr: boolean
+  copiedPrivKey: '' | 'loading' | 'done'
+}
+
+export class AddressBlock extends Component<
```
br4in3x, line 39: Could we please have this as a functional component?

br4in3x (commented):
This is great, thanks Nam! One last request

src/features/recieve/AddressBlock.tsx
namnm: Migrated to functional component

br4in3x (approved):

PR #21, opened 2021-04-11
Auto refresh rpc every 10s

namnm:


Dicklesworthstone (approved):

src/legacy/Routes.tsx
```diff
@@ -66,29 +66,29 @@ class RouteApp extends React.Component<any, any> {
...
-        this.setDisconnected,
-      )
-    } // Read the address book
-
-    ;(async () => {
-      const addressBook = await AddressbookImpl.readAddressBook()
+    // Auto refresh every 10s
+    this.rpcRefreshIntervalId = window.setInterval(() => {
+      rpc.refresh()
+    }, 10000)
```
br4in3x, line 86: This is already done by [rpc.configure ](https://github.com/pastelnetwork/pastel-electron-wallet/blob/master/src/legacy/rpc.ts#L47-L62)function, which is [being passed into Loading screen](https://github.com/pastelnetwork/pastel-electron-wallet/blob/9f333f1e3a7288b8e2278340471623cdcfe1aefb/src/legacy/Routes.tsx#L575-L577) and executed there.  Please comment call to refresh in the `configure` function [here](https://github.com/pastelnetwork/pastel-electron-wallet/blob/master/src/legacy/rpc.ts#L50-L52) like it's already done for `getPslPrice`, otherwise refresh will be called too often.

br4in3x (changes requested):

src/legacy/Routes.tsx
namnm: No, those calls are `setTimeout` which are called only once. This PR is about refresh every 10s, which uses `setInterval`

br4in3x: Oh, that's true. My bad

br4in3x (approved):

PR #22, opened 2021-04-11
Check sha56 pastel params

namnm:
- Check sha256 pastel params
- Add retry button to click on if failed


src/legacy/components/LoadingScreen.tsx
```diff
@@ -152,60 +153,67 @@ class LoadingScreen extends Component<any, any> {
...
         }
       }
+      return true
+    } catch (err) {
+      this.setState({
+        currentStatus: `Error downloading params. The error was: ${err}`,
+        errorEnsurePastelParams: true,
+      })
+      return false
     }
```
br4in3x, line 216: Got an error trying to run this PR (and the loading screen stuck on "Checking sapling-output.params..." step):
image
Probably related to all functions with `Sync` postfix. We really should use their async versions

br4in3x (changes requested):

br4in3x:
As mentioned in another conversation, let's have SHA-256 instead of md5. A good example of that can be found [here](https://github.com/so-ta/sha256-file/blob/master/index.js), or perhaps we can even use that package (it's not very popular, so I would think twice).

src/legacy/components/LoadingScreen.tsx
namnm: This error is on master as well, so it's not caused by this PR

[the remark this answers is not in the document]
namnm: What do you want to test? The package sha256-file already has test cases it self. Im not sure what detail you want to add a unit test here.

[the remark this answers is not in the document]
br4in3x: It should test if the function actually works, e.g. downloads files if the existing files are incorrect. I can make an example if it would help

br4in3x: Sorry, ignore that. For some reason it worked when I ran it again

br4in3x (approved):

PR #25, opened 2021-04-13
Add in a tab for in-app "lite" image editing (e.g., cropping, masking, resampling, etc.) using locally cached version of https://www.photopea.com

ngvtuan:
##### Commits
- [x] Add in a tab for in-app "lite" image editing (e.g., cropping, masking, resampling, etc.) using locally cached version of https://www.photopea.com
- [x] Add loading component

##### FreedCamp
- [x] [Add in a tab for in-app "lite" image editing (e.g., cropping, masking, resampling, etc.) using locally cached version of https://www.photopea.com] (https://freedcamp.com/dashboard/home/panel/task/38401307)


src/features/pastelLoading/PastelLoading.module.css
```diff
@@ -0,0 +1,81 @@
+.loading {
```
br4in3x, line 1: I think we can get away by using one of loaders from http://jonjaques.github.io/react-loaders/ - the "ball beat" type looks the closest to what you made. Reasons to not have our own custom loader for now is:
1. It's generic - doesn't specified in our design and can be anything
2. Increases complexity - we don't need to write a generic loader if there is an established library of generic loaders with 0 dependencies

src/features/pastelPhotopea/PastelPhotopeaModal.tsx
```diff
@@ -0,0 +1,43 @@
...
+        >
+          X
+        </button>
+        <div className={styles.iframe}>
+          <PastelLoading loadingIsOpen={isLoading} />
+          <iframe
+            id='photopea'
+            src='https://www.photopea.com'
+            onLoad={() => setTimeout(() => setIsLoading(false), 100)}
+          />
```
br4in3x, line 38: Can we have this preloaded? It takes significant time to show up. Here is an example:
https://codepen.io/darknessHO/pen/goNvdK. I think we can use [React Portals](https://reactjs.org/docs/portals.html) for that. We can have this iframe load and be invisible and render it in this component through a portal. 

This is just an idea, perhaps you can come up with your own mechanism :)

src/features/pastelPhotopea/PastelPhotopeaModal.tsx
```diff
@@ -0,0 +1,43 @@
...
+          onClick={() => dispatch(closePastelPhotopeaModal())}
+        >
+          X
+        </button>
+        <div className={styles.iframe}>
+          <PastelLoading loadingIsOpen={isLoading} />
+          <iframe
+            id='photopea'
+            src='https://www.photopea.com'
+            onLoad={() => setTimeout(() => setIsLoading(false), 100)}
```
br4in3x, line 37: Why do we need `setTimeout` here?

src/features/pastelPhotopea/PastelPhotopeaModal.tsx
```diff
@@ -0,0 +1,43 @@
...
+import styles from './PastelPhotopeaModal.module.css'
+import { closePastelPhotopeaModal } from './PastelPhotopeaModalSlice'
+
+export default function PastelPhotopeaModal(): JSX.Element {
+  const { modalIsOpen } = useAppSelector(state => state.pastelPhotopeaModal)
+  const dispatch = useAppDispatch()
+  const [isLoading, setIsLoading] = useState(true)
+
+  if (!modalIsOpen) {
+    return <></>
```
br4in3x, line 15: Perhaps `return null` is better?

br4in3x (changes requested):

src/features/pastelPhotopea/PastelPhotopeaModal.tsx
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  return (
+    <Modal
```
br4in3x, line 19: When the modal opens, it has a large white border. Let's remove the border to provide more space for the iframe.

src/features/pastelLoading/PastelLoading.module.css
ngvtuan: @br4in3x I have updaed for the loading by removed it

src/features/pastelPhotopea/PastelPhotopeaModal.tsx
ngvtuan: yes we can i have updated the code

src/features/pastelPhotopea/PastelPhotopeaModal.tsx
ngvtuan: Updated

src/features/pastelPhotopea/PastelPhotopeaModal.tsx
ngvtuan: Updated

br4in3x (approved):

br4in3x (approved):

PR #29, opened 2021-04-20
add pastel-rpc: network-stats && types

Dinisueldanov:
### Summary
- Added global types.ts for network-stats to src.
- Added network-stats in api/pastel-rpc.
- Imported sql.js for Database type.


package.json
```diff
@@ -63,6 +63,7 @@
     "regenerator-runtime": "0.13.7",
     "request": "2.88.2",
     "sha256-file": "1.0.0",
+    "sql.js": "^1.5.0",
```
namnm, line 66: Please note all other packages are exact match (means no `^`) can you remove it and reinstall to update yarn.lock? You can use `-E` flag in the yarn command `yarn add -E sql.js`.

package.json
```diff
@@ -98,6 +99,7 @@
     "@types/request": "2.48.5",
     "@types/sha256-file": "1.0.0",
     "@types/source-map-support": "0.5.3",
+    "@types/sql.js": "^1.4.2",
```
namnm, line 102: Same with above

src/type.ts
```diff
@@ -0,0 +1,205 @@
+/* eslint-disable */
```
namnm, line 1: Why do you disable eslint on this file? Please remove it and let the lint run.

namnm (changes requested):

src/type.ts
Dinisueldanov: Thanks for your review.
For data types that have not yet been determined, any type was used.
It caused an eslint warning.

namnm (approved):

src/type.ts
br4in3x: @Dinisueldanov we shouldn't disable eslint for anything actually, if you have to do it means something is going wrong. Please find a way to fix the issue without disabling esLint

src/type.ts
```diff
@@ -0,0 +1,205 @@
...
+import { Database } from 'sql.js'
```
br4in3x, line 2: As far as I understand, this file is only used in the `pastel-rpc/network-stats` folder. Please move it there.

src/type.ts
```diff
@@ -0,0 +1,205 @@
...
+
+export interface IRawMempoolInfo {
+  transactionid: string
+  size: number
+  fee: number
+  time: number
+  height: number
+  startingpriority: number
+  currentpriority: number
+  depends: any
```
br4in3x, line 49: We should not have any `any` types in our code :) It defeats the purpose of TypeScript. Please change it to the actual type everywhere in this PR

src/api/pastel-rpc/network-stats/network-stats.ts
```diff
@@ -0,0 +1,480 @@
...
+      hashrate: hashrate,
+      difficulty: difficulty,
+    }
+  } catch (error) {
+    throw new Error(`network-stats getStatisticInfo error: ${error.message}`)
+  }
+}
+
+export async function getNetworkHashps(config: TRPCConfig): Promise<string> {
+  const { result } = await rpc('getnetworkhashps', [], config)
```
br4in3x, line 43: Could we please change it to `rpc<T>(...)`, where T is the response type of this RPC call for all functions in this file? Make sure you are returning the right type though, because I know that `getnetworkhashps` returns this object as response:
```
{
  "result": 14523646,
  "error": null,
  "id": "curltest"
}
```
Notice that "result" here is a number, not a string

br4in3x: To properly check what is the return type of every call, you can do the following:
1. Download Insimnia (if you don't have it yet): https://insomnia.rest/
2. Import my config (I created the first 3 methods): https://gist.github.com/br4in3x/a16e37f0040829c3db5190a0b2198ce3
3. Run `pasteld`:
  3.1. cd into the folder with binaries: `cd static/bin/<your system>`
  3.2. run pasteld rpc server: `.\pasteld-win.exe --rpcport=9932 --rpcuser=insomnia --rpcpassword=insomnia`
4. Execute any call from insomnia, you will get something like this:
image

package.json
```diff
@@ -98,6 +99,7 @@
     "@types/request": "2.48.5",
     "@types/sha256-file": "1.0.0",
     "@types/source-map-support": "0.5.3",
+    "@types/sql.js": "1.4.2",
```
br4in3x, line 102: Let's have this changes in the SQL PR please

src/type.ts
```diff
@@ -0,0 +1,464 @@
+export interface INetworks {
```
br4in3x, line 1: Please move these types into the network-stats feature folder since they are only used there

package.json
```diff
@@ -63,6 +63,7 @@
     "regenerator-runtime": "0.13.7",
     "request": "2.88.2",
     "sha256-file": "1.0.0",
+    "sql.js": "1.5.0",
```
br4in3x, line 66: Let's not include this dependency in this PR (we're not using it here)

yarn.lock
```diff
@@ -10415,6 +10415,11 @@ sprintf-js@~1.0.2:
   resolved "https://registry.yarnpkg.com/sprintf-js/-/sprintf-js-1.0.3.tgz#04e6926f662895354f3dd015203633b857297e2c"
   integrity sha1-BOaSb2YolTVPPdAVIDYzuFcpfiw=
 
+sql.js@1.5.0:
+  version "1.5.0"
+  resolved "https://registry.yarnpkg.com/sql.js/-/sql.js-1.5.0.tgz#6625c3e41115194f6bfab80a5e53647f1378c3fb"
+  integrity sha512-Qqr6HgX/hCDpLFWdN0BNoNpYQ2c1tOl1c3HGI0cshjaFSAWszKICuLZ9CyFUvRFPpEGW8RzHzwuXWWvXVGTKBg==
+
```
br4in3x, line 10422: Please see my previous comment

br4in3x (approved):

PR #33, opened 2021-04-23
Pastel sqlite/db

Dinisueldanov:
### Summary
- Added the pastel-sqlite db feature(create && open && save).
- Added the global redux store for pastel-sqlite db.
- Fixed memory-leak issue when app is closed


src/features/pastelDB/pastelDBLib.ts
```diff
@@ -0,0 +1,80 @@
...
+  create_rawmempoolinfo,
+  create_rawtransaction,
+  create_statisticinfo,
+  create_totalbalance,
+  create_transaction,
+  create_txoutsetinfo,
+  create_walletinfo,
+} from './constants'
+
+const initSqlJs = require('./sql-wasm.js')
```
namnm, line 26: Why dont we use import here, but need to use require?

.eslintrc.js
```diff
@@ -39,6 +39,7 @@ module.exports = {
         },
       },
     ],
+    '@typescript-eslint/no-var-requires': 'off',
```
namnm, line 42: If you fix the below comment, you dont need to turn off this rule

package.json
```diff
@@ -100,12 +103,14 @@
...
     "@types/sha256-file": "1.0.0",
     "@types/source-map-support": "0.5.3",
+    "@types/sql.js": "1.4.2",
     "@types/underscore": "1.11.0",
     "@types/uuid": "8.3.0",
     "@types/ws": "7.4.0",
     "@typescript-eslint/eslint-plugin": "4.0.1",
     "@typescript-eslint/parser": "4.20.0",
     "babel-loader": "8.2.2",
+    "copy-webpack-plugin": "^6.3.2",
```
namnm, line 113: Here we have `^`, please remove it and reinstall to update `yarn.lock`

src/features/pastelDB/constants/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './common'
+export * from './querys'
```
namnm, line 2: Is this typo, should be `queries`?

src/features/pastelDB/pastelDBLib.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+  const newdb: Database = new SQL.Database(filebuffer)
+  return newdb
+}
+
+export const saveDataToLocalSqlite = async (db: Database): Promise<void> => {
+  const data = db.export()
+  const buffer = Buffer.from(data)
+  fs.writeFileSync(
+    path.join(process.env.HOMEPATH as string, './pasteldb.sqlite'),
```
namnm, line 53: Can we move this to a separated file, like `src/config.ts`?

namnm: Also there would be a APP_DATA path of this app already, you should use it instead. For eg: see `src/legacy/components/LoadingScreen.tsx`

src/features/pastelDB/pastelDBLib.ts
```diff
@@ -0,0 +1,80 @@
...
+  const SQL = await initSqlJs({
+    locateFile: (file: string) => {
+      return `static/bin/${file}`
+    },
+  })
+
+  let filebuffer
+  try {
+    filebuffer = fs.readFileSync(
+      path.join(process.env.HOMEPATH as string, './pasteldb.sqlite'),
```
namnm, line 38: Can we move this to a separated file, like src/config.ts?
Also there would be a APP_DATA path of this app already, you should use it instead. For eg: see src/legacy/components/LoadingScreen.tsx

package.json
```diff
@@ -100,12 +103,14 @@
     "@types/request": "2.48.5",
     "@types/sha256-file": "1.0.0",
     "@types/source-map-support": "0.5.3",
+    "@types/sql.js": "1.4.2",
```
br4in3x, line 106: Fixed sql.js problem, please see the solution here: https://github.com/pastelnetwork/pastel-electron-wallet/compare/pastel-sqlite/db...pastel-sqlite/db-fix

src/features/pastelDB/pastelDBLib.ts
br4in3x, line 53: We should use pastel local settings directory, see https://github.com/pastelnetwork/pastel-electron-wallet/blob/master/src/legacy/components/LoadingScreen.tsx#L34-L44

br4in3x (changes requested):
Please address all comments in this PR and also copy changes from https://github.com/pastelnetwork/pastel-electron-wallet/compare/pastel-sqlite/db...pastel-sqlite/db-fix

src/features/pastelDB/sql-wasm.js
```diff
@@ -0,0 +1,3094 @@
+// We are modularizing this manually because the current modularize setting in Emscripten has some issues:
```
namnm, line 1: This file and the below `wasm` file are like binaries not code, do you think it's better to put it in `src/assets/wasm/` or something?

namnm (changes requested):

src/features/pastelDB/sql-wasm.js
br4in3x: Fixed this on a separate branch for @Dinisueldanov to copy changes from: https://github.com/pastelnetwork/pastel-electron-wallet/blob/eb085168be7ee800fc99e83993dc13cd97e66f8e/package.json#L68

src/legacy/components/LoadingScreen.tsx
```diff
@@ -247,23 +247,19 @@ class LoadingScreen extends Component<any, any> {
     this.loadPastelConf(false)
```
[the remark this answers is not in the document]
Dinisueldanov: Thanks for your review.
Actually I removed following code from the appquitting event's callback.
Due to this code, the memory-leak issue occurred and didn't also work this code when app is close.

`
const { rpcConfig } = this.state

//removed this line
this.setState({
  currentStatus: 'Waiting for pasteld to exit...',
})

history.push(routes.LOADING)
`

Would you like to return the code to its original state even if there is a issue?

package.json
Dinisueldanov: Thanks for the comment.
I updated my code according to your comment and sql-wasm.js issue is fixed perfectly but if I couldn't install @types/sql.js and type error occurs still.

`
import initSqlJs, { Database } from 'sql.js'
Could not find a declaration file for module 'sql.js'. 'D:/git/pastel-electron-wallet/node_modules/sql.js/dist/sql-wasm.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/sql.js` if it exists or add a new declaration (.d.ts) file containing `declare module 'sql.js';`
`

src/legacy/components/LoadingScreen.tsx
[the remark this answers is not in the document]
Dinisueldanov: Thanks for your comment.
Actually, I've removed following code from the appquitting event callback.
This code made the memory-leak error when app is close and component would be unmounted and didn't work.

`
const { rpcConfig } = this.state

//removed this line by me
this.setState({
  currentStatus: 'Waiting for pasteld to exit...',
})

history.push(routes.LOADING)
`

if you want to back to origin code even if there is error, I will do.

src/config.ts
```diff
@@ -0,0 +1,17 @@
...
+  )
+}
+
+export const writeSqliteDBFile = async (buffer: Buffer): Promise<void> => {
+  await fs.promises.writeFile(
+    path.join(remote.app.getPath('appData'), 'Pastel', 'pasteldb.sqlite'),
+    buffer,
+    { flag: 'a+' },
+  )
+}
```
br4in3x, line 17: This file should only contain constants like `PastelConfigFolder`, and not functions. Functions are a part of DB feature, and should live in that folder

src/legacy/components/LoadingScreen.tsx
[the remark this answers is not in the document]
br4in3x: Okay, if this fixes a bug in legacy, let's leave it

src/legacy/containers/Root.tsx
```diff
@@ -4,13 +4,15 @@ import { MemoryRouter } from 'react-router-dom'
...
 
-const Root = (): JSX.Element => (
-  <>
-    <MemoryRouter>
-      <Routes />
-    </MemoryRouter>
-    <ReduxErrorModal />
-  </>
-)
+const Root = (): JSX.Element => {
```
br4in3x, line 7: I don't think we need this change

package.json
```diff
@@ -40,6 +40,7 @@
     "electron-log": "4.3.2",
     "electron-squirrel-startup": "1.0.0",
     "electron-store": "5.1.1",
+    "fs-extra": "9.1.0",
```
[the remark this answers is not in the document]
br4in3x: This wasn't addressed

package.json
```diff
@@ -84,6 +86,7 @@
     "@types/classnames": "2.2.11",
     "@types/dateformat": "3.0.1",
     "@types/electron-devtools-installer": "2.2.0",
+    "@types/fs-extra": "9.0.11",
```
br4in3x, line 89: We don't need this package

br4in3x (commented):
Nice work, please run `yarn` after any changes to the `package.json` file. This would update `yarn.lock` as well.

package.json
Dinisueldanov: Thanks for your comment.
We need to use this "@types/fs-extra": "9.0.11" package. 
Now we are using fs-extra. I tried to use fs but it doesn't work. it makes the malformed error. And they recommended using fs-extra. and you know, "@types/fs-extra": "9.0.11" package didn't be provided as a default, so we need it.
I tried to remove this package but got this error message.
```
Could not find a declaration file for module 'fs-extra'. 'D:/git/pastel-electron-wallet/node_modules/fs-extra/lib/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/fs-extra` if it exists or add a new declaration (.d.ts) file containing `declare module 'fs-extra';`
```
Please check it again and let me know your opinion.
Thanks

namnm: > I tried to use fs but it doesn't work

What is the detail error of usage of original `fs`?

Dinisueldanov: I got following error when I use fs and keep the db data.

```
await fs.promises.writeFile(
    path.join(remote.app.getPath('appData'), 'Pastel', 'pasteldb.sqlite'),
    buffer,
    { flag: 'a+' },
  )
```
`This file is malformed.`

I think this file is corrupt, but fs-extra does work well.

Dinisueldanov:
- Opened new PR#43 instead of this PR.

PR #34, opened 2021-04-24
Retry params download upto 10 times

namnm:


src/features/loading/utils.ts
```diff
@@ -108,7 +108,5 @@ export const checkHashAndDownloadParams = async ({
     promises.push(promise)
   }
 
-  for (const promise in promises) {
-    await promise
-  }
+  await Promise.all(promises)
```
br4in3x, line 111: This will make to run the check for every file in parallel, making calls to onUpdate callback without order, resulting to status message on the loading screen to change chaotically

src/legacy/components/LoadingScreen.tsx
```diff
@@ -144,6 +144,9 @@ class LoadingScreen extends Component<any, any> {
       })
       return true
     } catch (err) {
+      if (nRetry < 10) {
```
br4in3x, line 147: I think this will make the ensurePastelParams to run 10 times even if all parameters are valid.

src/features/loading/utils.ts
namnm: No:
- First: `for (const promise in promises) {` the variable `promise` here is a string, because you were using `for ... in ...` you should use `for ... of ...` instead
- Second: At the time you initialize the promise in the above bigger `for` loop, it start checking/downloading immediately. So if you want that logic, you need to await inside the above bigger `for` loop like the old logic.

src/legacy/components/LoadingScreen.tsx
namnm: Why? If they are valid, they will will never run into this `catch` it will be the above `return true`

src/features/loading/utils.ts
namnm: @br4in3x Here I made an example for you:
- `forInLoop` is your logic, it would run in ~0ms because await on string is noop.
- `forOfLoop` is the fixed version, it would run in ~1000ms (the longest execution of all promises). It is equivalent to the `Promise.all` version.
- `oldForLoop` I guess it is what you want, which waits until the previous promise resolved then execute the next. It would run in ~2000ms

Screen Shot 2021-04-26 at 02 07 20

namnm:
Merged on https://github.com/pastelnetwork/pastel-electron-wallet/pull/37

br4in3x (approved):

src/features/loading/utils.ts
br4in3x: Addressed in https://github.com/pastelnetwork/pastel-electron-wallet/pull/37

PR #35, opened 2021-04-25
Expert Console additional features

westalert:
update for command


Dicklesworthstone (approved):

namnm:
@Dicklesworthstone Tested several times more. PR should be ok now. Changes I made:
- Fix css to allow scroll, I think he made it scrollable but in this new PR, scroll doesnt work
- Fix css to allow copy text/command in the terminal output
- Fix bug display/format array from rpc response
- Added 2 new logic when user press tab: Nothing typed: show help | Multiple matches: show list of matches
- Improve scroll experience and clickable area of the input, when the terminal width is too big
- Remove unused files from old library from code base, he switched to a new library
- Fix all eslint errors to pass CI

br4in3x (dismissed):
Reviewing with a minimal effort because this would be refactored anyway. @westalert

namnm (approved):

src/features/pastelExpertConsole/rpc-services/api.ts
```diff
@@ -0,0 +1,71 @@
...
+            reject({
+              message,
+              statusCode: statusCode,
+            })
+          })
+      })
+    },
+  }),
+  apiMethods,
+)
```
br4in3x, line 69: We already have an RPC client. Please see if you can integrate our client into your function. Here is our client: https://github.com/pastelnetwork/pastel-electron-wallet/blob/master/src/api/pastel-rpc/rpc.ts#L9

src/legacy/components/Receive.tsx
```diff
@@ -1,369 +0,0 @@
-/* eslint-disable */
```
br4in3x, line 1: I don't think this file should be removed. Could you provide a reason?

namnm: @br4in3x The PR with receive in the past, didnt remove this file. I checked and there's no reference to this file. His PR accidentally re-introduced `Receive.css.module`, when I removed that file, I noticed this and also removed this as well.

src/features/pastelExpertConsole/rpc-services/api.ts
```diff
@@ -0,0 +1,71 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
br4in3x, line 1: Please do not disable this rule

src/features/pastelExpertConsole/rpc-services/utils.ts
```diff
@@ -0,0 +1,979 @@
...
+  'z_sendmany',
+  'z_setmigration',
+  'z_shieldcoinbase',
+  'z_viewtransaction',
+  'zcbenchmark',
+  'zcrawjoinsplit',
+  'zcrawkeygen',
+  'zcrawreceive',
+  'zcsamplejoinsplit',
+]
```
br4in3x, line 155: This allows to use this as a type
```suggestion
] as const
```

src/features/pastelExpertConsole/rpc-services/utils.ts
```diff
@@ -0,0 +1,979 @@
...
+
+export type APIMethods = {
```
br4in3x, line 157: After we change `METHODS` to be with `as const`, this would be broken down into separate response types

src/features/pastelExpertConsole/rpc-services/api.ts
```diff
@@ -0,0 +1,71 @@
...
+      return 'Something went wrong'
+  }
+}
+
+const apiMethods: any = {}
+
+const api: APIMethods = METHODS.reduce(
+  (obj, method) => ({
+    ...obj,
+    [method]: (params = []) => {
```
br4in3x, line 32: Using `reduce` and spread operator will create a copy of the `apiMethods` object as many times as there are elements in the `METHODS` array, which is inefficient. Instead, we should use good old `forEarch` or even `for`.

src/features/pastelExpertConsole/rpc-services/api.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  switch (statusCode) {
+    case 401:
+      return 'Not authorized to access RPC, please check your rpcuser and rpcpassword'
+    default:
+      return 'Something went wrong'
+  }
+}
+
+const apiMethods: any = {}
```
br4in3x, line 27: Please think about a type that is better than "any". This could be:
```ts
type possibleMethods = typeof METHODS[number]
type dynamicAPI = {
  [x in possibleMethods]: <T>(params: string[]) => Promise<T>
}

const apiMethods = {} as dynamicAPI
```
And of course, you can select your own names :)

Now, we have a fully typed `apiMethods` object, that can be called like that:
```ts
const r = apiMethods('z_importwallet')<TGetBlockResult>(['hash']) // TGetBlockResult is the new type from broken down APIMethods type. 
```
The `r` would become a type of `TGetBlockResult`.

br4in3x (approved):

br4in3x (approved):

br4in3x (approved):

PR #36, opened 2021-04-25
Add: full pastel-rpc/sqlite db features

Dinisueldanov:
### Summary
- Fetch real-time statistics data.
- Store Fetched data to Redux DB.
- Keep Redux DB data to local file when app is closed.


br4in3x:
This PR is too big and includes changes from https://github.com/pastelnetwork/pastel-electron-wallet/pull/33.

PR #39, opened 2021-04-25
Adds eslint-plugin-react

br4in3x:
This PR adds eslint-plugin-react package to enforce best practices for writing React components.


namnm:
@br4in3x CI failed. Btw, should we use eslint config airbnb? It is proven the best of all config out there.

PR #43, opened 2021-04-28
Pastel sqlite/db no redux

Dinisueldanov:
### Summary
- Removed redux for db instance.
- Node js Singleton pattern for db instance.


src/main.ts
```diff
@@ -157,3 +165,7 @@ app.on('activate', () => {
     createWindow()
   }
 })
+
+app.on('ready', async () => {
+  global.pastelDB = await createDatabase()
```
br4in3x, line 170: Please handle the exception with try/catch

br4in3x, line 170: I don't think we need global to be honest. We can handle this with NodeJS Singleton pattern. You can create a new file, create a new object, and export it. Then you can import it, and assign "database" property on it. Next time when you import that object, it'll be pointing to this DB instance. Like this:
```js
// database.ts
export default {
  database: Database | null = null
}
```

```ts
// main.ts
import d from './database.js'
d.database = await createDatabase();
```

```ts
// anywhere
import { database } from './database.js'
// database is now our created database
```

Here is the article about it: https://medium.com/swlh/node-js-and-singleton-pattern-7b08d11c726a

src/features/pastelDB/pastelDBLib.ts
```diff
@@ -0,0 +1,111 @@
...
+    )) === false
+  ) {
+    await sleep(1000)
+  }
+
+  const SQL = await initSqlJs({
+    locateFile: (file: string) => {
+      return `${app.getAppPath()}/.webpack/renderer/static/bin/${file}`
+    },
+  })
```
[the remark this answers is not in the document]
Dinisueldanov: Thanks for your comment.
Actually, origin path(static/bin/${file}) worked well on react component. appointed build app full path 
But, when it is called this on main.ts, (static/bin/${file}) did not work correctly, only appoint 'static/bin/pastelsqlite.db'.
So, I needed to use the full path.

src/main.ts
Dinisueldanov: Good advise, Thank you.

Dinisueldanov: Thanks for you good solution,
Will update with this.

src/index.tsx
```diff
@@ -19,6 +20,12 @@ setInterval(() => {
   store.dispatch(fetchPastelPrice())
 }, oneHour)
 
+try {
+  PastelDB.getDatabaseInstance()
+} catch (error) {
+  throw new Error('"pastelDB create:" error')
```
br4in3x, line 26: If you re-throw the error here, on reason to use try/catch. The reason for it is to allow users to use the wallet even if there are problems creating DB
```suggestion
  // TODO log errors to a central logger so we can address them later.
  console.error(`PastelDB.getDatabaseInstance error: ${error.message}`)
```

br4in3x (commented):
Thanks for the changes :) One last thing

src/index.tsx
br4in3x: Could add the a `// TODO` as I asked please? If you check our codebase, this comment precedes all `console` statements

br4in3x (commented):
Please see my comment here https://github.com/pastelnetwork/pastel-electron-wallet/pull/43#discussion_r625153693

src/index.tsx
Dinisueldanov: okay, added.

br4in3x (approved):

PR #44, opened 2021-04-28
Pastel sqlite/db full functions

Dinisueldanov:
### Summary
- Added db operation functions.
- Added some types.


src/features/pastelDB/pastelDBLib.ts
```diff
@@ -109,3 +125,408 @@ export const createTables = async (db: Database): Promise<void> => {
   db.exec(create_listaddresses)
   return
 }
+
```
br4in3x, line 128: Could you try to add tests to this new functions, please? They seems to be a perfect candidate. Here is an example: https://github.com/pastelnetwork/pastel-electron-wallet/blob/master/src/features/loading/__tests__/utils.test.ts

br4in3x (changes requested):
Please add tests

src/features/pastelDB/pastelDBLib.ts
```diff
@@ -63,23 +96,463 @@ export const saveDataToLocalSqlite = async (db: Database): Promise<void> => {
...
+export const validateDataFromDB = (
+  pastelDB: Database,
+  tableName: string,
+  transactionid: string,
+  time: number,
+): boolean => {
+  const sqlText = `SELECT id FROM ${tableName} WHERE transactionid=$tid AND time=$time`
```
br4in3x, line 126: Let's move this to queries (and all other SQL queries to queries.ts file).

Also, we need to add some table name validations if `$tableName` is not available as a query parameter. The reason I'm asking is to guard us from SQL injections: imagine I provide this string instead of `$tablename`: `(DROP TABLE blockinfo);`. 

Example of validations:
```
const tableNames = {
  blockinfo: true,
  transaction_tbl: true,
  // ... other tables
}
if (tableNames[tableName] !== true) {
  throw new Error(`pastelID validateDataFromDB error: table name is invalid`)
}
```

src/features/pastelDB/pastelDBLib.ts
Dinisueldanov: - Added tests

src/features/pastelDB/pastelDBLib.ts
Dinisueldanov: - Good comments, updated.

Dinisueldanov:
> 
> 
> Please add tests

- Added tests here.

src/features/pastelDB/pastelDBLib.ts
```diff
@@ -132,18 +128,32 @@ export const validateDataFromDB = (
...
-  const sqlText = `SELECT id FROM  ${tableName}`
+  if (tableNames[tableName] !== true) {
+    throw new Error('pastelDB getLastIdFromDB error: table name is invalid')
+  }
+
+  const sqlText = selectIDQuery + tableName + orderByIDQuery
   const sqlResult = pastelDB.exec(sqlText)
   if (sqlResult.length) {
-    return sqlResult[0].values.length + 1
+    return sqlResult[0].values[0][0] as number
```
br4in3x, line 156: `as` is generally evil. Let's convert this with `parseInt`

src/features/pastelDB/pastelDBLib.ts
```diff
@@ -154,78 +164,89 @@ export const insertStatisticDataToDB = (
...
+
+  pastelDB.exec(insertNetworkinfoQuery, values)
 }
 
 export const insertNetTotalsToDB = (
   pastelDB: Database,
   nettotals: TNetTotals,
 ): void => {
-  const create_timestamp: string = new Date().toLocaleTimeString()
+  const createTimestamp: string = new Date().toLocaleTimeString()
```
br4in3x, line 222: Why is this createTimestamp const using `toLocaleTimeString()` and others are not?

src/features/pastelDB/__tests__/pastelDBLib.test.ts
```diff
@@ -0,0 +1,488 @@
+import initSqlJs, { Database, QueryExecResult } from 'sql.js'
```
br4in3x, line 1: Can we mock imports, so it's actually testing only the code in queries, and nothing else? Please see the example here: https://github.com/pastelnetwork/pastel-electron-wallet/blob/master/src/features/loading/__tests__/utils.test.ts. Here is the article about jest mocks: https://jestjs.io/docs/manual-mocks

br4in3x (changes requested):

src/features/pastelDB/__tests__/pastelDBLib.test.ts
```diff
@@ -0,0 +1,488 @@
...
+    db.exec(createTransaction)
+    db.exec(createTxoutsetinfo)
+    db.exec(createChaintips)
+    db.exec(createBlocksubsidy)
+    db.exec(createWalletinfo)
+    db.exec(createListreceivedbyaddress)
+    db.exec(createListtransactions)
+    db.exec(createListunspent)
+    db.exec(createTotalbalance)
+    db.exec(createListaddresses)
```
br4in3x, line 69: What is the purpose of this? If this is to initialize database structure, then it should be placed in `beforeAll`: https://jestjs.io/docs/setup-teardown

src/features/pastelDB/__tests__/pastelDBLib.test.ts
```diff
@@ -0,0 +1,488 @@
...
+
+    const values = {
+      $newId: 1,
+      $hashrate: 1.0,
+      $difficulty: 1.2345678,
+      $createTimestamp: 0,
+    }
+
+    const result: QueryExecResult[] = db.exec(insertStatisticinfoQuery, values)
+    expect(result).not.toBeNull()
```
br4in3x, line 87: Let's compare the result with expected value, not just `null`.

br4in3x (changes requested):

src/features/pastelDB/__tests__/pastelDBLib.test.ts
Dinisueldanov: Good comment, thank you.

br4in3x (approved):

PR #45, opened 2021-04-29
Pastel sqlite/db full manage

Dinisueldanov:
### Summary

- Added features that fetch data and save to virtual DB in real time.


src/features/pastelDB/pastelDBThread.tsx
```diff
@@ -0,0 +1,227 @@
+import { remote } from 'electron'
```
br4in3x, line 1: Let's try to add tests here. An example is here: https://github.com/pastelnetwork/pastel-electron-wallet/blob/master/src/features/loading/__tests__/utils.test.ts

br4in3x:
Going to review this after https://github.com/pastelnetwork/pastel-electron-wallet/pull/44 gets merged.

src/features/pastelDB/pastelDBThread.tsx
Dinisueldanov: - Added tests.

src/features/pastelDB/pastelDBThread.tsx
```diff
@@ -0,0 +1,227 @@
...
+}
+
+export const PastelDBThread = (
+  props: PastelDBThreadProps,
+): React.ReactElement => {
+  const { rpcConfig } = props
+  const [isStarted, setStarted] = useState(false)
+
+  useEffect(() => {
+    const pastelDB = remote.getGlobal('pastelDB')
```
br4in3x, line 182: I think we fixed this, so it's not global anymore (there was a `database` object)

src/features/pastelDB/pastelDBThread.tsx
```diff
@@ -0,0 +1,227 @@
...
+          fetchChaintips(pastelConfig),
+          fetchBlocksubsidy(pastelConfig),
+          fetchWalletinfo(pastelConfig),
+          fetchListTransactions(pastelConfig),
+          fetchListunspent(pastelConfig),
+          fetchTotalbalance(pastelConfig),
+          fetchListaddresses(pastelConfig),
+        ])
+
+        setTimeout(() => fetchStatisticDataFromRPC(), 10000)
```
br4in3x, line 211:
```ts
        // in the beggning of the function
        let timer: number = 0;

        // here
        timer = setTimeout(() => fetchStatisticDataFromRPC(), 10000)
```
But also, this is a bit strange to me. Why is this on the timeout? Can we have some event controlling it rather an arbitrary set amount of time?

src/features/pastelDB/pastelDBThread.tsx
```diff
@@ -0,0 +1,227 @@
...
+const fetchListaddresses = async (props: fetchFuncConfig) => {
+  const addresslist = await listAddresses(props.rpcConfig)
+  for (let i = 0; i < addresslist.length; i++) {
+    insertListaddresses(props.pastelDB, addresslist[i])
+  }
+}
+
+export const PastelDBThread = (
+  props: PastelDBThreadProps,
+): React.ReactElement => {
```
br4in3x, line 177: This shouldn't be a React component. React Components provide UI or provide logic for UI, and this function has nothing to do with UI at all. Therefore, it should be converted into a simple named JavaScript function that will be called on some kind of event

src/features/pastelDB/__tests__/pastelDBThread.test.ts
```diff
@@ -0,0 +1,473 @@
...
+  const listAddresses = jest.fn(() => {
+    return 'address1'
+  })
+
+  test('statistic data should fetch and store correctly', async () => {
+    const info = getStatisticInfo()
+    insertStatisticDataToDB(pastelDB.db, info.hashrate, info.difficult)
+    const result = getDatafromDB('statisticinfo')
+    expect(result[0].values[0]).toContainEqual('1.2345')
+  })
```
br4in3x, line 367: This is not right. Since database methods already tested in [pastelLib tests](https://github.com/pastelnetwork/pastel-electron-wallet/blob/42ba388af1f0a3fdf3d40165ef5498c2b98574a6/src/features/pastelDB/__tests__/pastelDBLib.test.ts), we can mock them because we can assume they work:

Here is how this test is supposed to look like:

```ts
import * as pastelDBLib from './pastelDBLib`
import * as netStats from '../../api/pastel-rpc/network-stats/network-stats' // after addressing all the comments, this should go away too

test('fetchStatisticinfo works', async () => {
  // Arrange
  const mockDB = { database: {} } as Database; // this is bad actually, but we don't have database interface at the moment. If we had, we could actually mock every field 
  const mockConfig = { username: 'test', password: 'pass' } // doesn't matter what's in there 
  const insertStatisticDataToDBSpy = jest.spyOn(pastelDBLib, 'insertStatisticDataToDBSpy')
  const getStatisticInfoSpy = jest.spyOn(netStats, 'getStatisticInfo') // this will be replaced with rpc method mock after my comments get addressed
  
  // Act
  const info = await fetchStatisticinfo({ pastelDB: mockDB, rpcConfig: mockConfig })

  // Assert
  expect(getStatisticInfoSpy).toHaveBeenCalled()
  expect(getStatisticInfoSpy).toHaveBeenCalledWith(mockConfig)

  expect(insertStatisticDataToDBSpy).toHaveBeenCalled()
  expect(insertStatisticDataToDBSpy).toHaveBeenCalledWith(mockDB, 10, 10)
})
```
Also please take a look at `Arrange`, `Act`, and `Assert` comments. They are good for visual separation of logic blocks. Please fix all your tests to look like this

src/api/pastel-rpc/network-stats/network-stats.ts
```diff
@@ -119,8 +121,8 @@ export async function getRawMempool(
 }
 
 export async function getBlockByHeight(config: TRPCConfig): Promise<string> {
-  // const height = 1000
-  const { result } = await rpc<TGetblockhash>('getblockhash', [], config)
+  const height = 1000
+  const { result } = await rpc<TGetblockhash>('getblockhash', [height], config)
   return result
 }
```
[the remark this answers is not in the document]
Dinisueldanov: Thanks for your comment.
Done.

src/features/pastelDB/pastelDBThread.tsx
Dinisueldanov: Now we are fetching the data from RPC periodically for statistics data.
Do you mean that should fetch the data by some event? Actually, we don't have any event for this.

src/features/pastelDB/pastelDBThread.tsx
```diff
@@ -96,132 +117,201 @@ const fetchRawMempoolinfo = async (props: fetchFuncConfig) => {
...
-          fetchListTransactions(pastelConfig),
-          fetchListunspent(pastelConfig),
-          fetchTotalbalance(pastelConfig),
-          fetchListaddresses(pastelConfig),
-        ])
-
-        setTimeout(() => fetchStatisticDataFromRPC(), 10000)
-      } else {
-        return
+export async function PastelDBThread(rpcConfig: RPCConfig): Promise<void> {
```
br4in3x, line 272: This is a pretty important function. Can we have a test for it as well?

src/features/pastelDB/__tests__/pastelDBThread.test.ts
```diff
@@ -295,179 +152,380 @@ describe('PastelDBThread', () => {
...
+      .mockImplementation()
+
+    // Act
+    await pastelDBThread.fetchListaddresses({
+      pastelDB: pastelDB.db,
+      rpcConfig: mockConfig,
+    })
+
+    // Assert
+    expect(rpcSpy).toHaveBeenCalled()
```
br4in3x, line 525: If there is `toHaveBeenCalledWith` in your test, `toHaveBeenCalled` is not required anymore

br4in3x (commented):
Almost there!

br4in3x (dismissed):

src/api/pastel-rpc/network-stats/network-stats.ts
```diff
@@ -0,0 +1,313 @@
...
+  TRawTransaction,
+  TTotalBalance,
+  TTransactionInfo,
+  TTxout,
+  TTxoutsetInfo,
+  TValidateaddress,
+  TValidateaddresses,
+  TWalletInfo,
+  TZListReceivedByAddress,
+} from './type'
```
br4in3x: I think this should be `import * as type from './type'`

src/api/pastel-rpc/network-stats/network-stats.ts
```diff
@@ -0,0 +1,313 @@
...
+
+type TStatistic = {
+  hashrate: number
+  difficulty: number
+}
+
+export async function getStatisticInfo(
```
br4in3x: All functions in this file are just wrappers of the `rpc` method. Please remove them

src/api/pastel-rpc/rpc.ts
```diff
@@ -6,14 +6,15 @@ export type TRPCConfig = {
   password: string
 }
 
+export type TRPCParam = string | boolean | number
```
br4in3x: PLease remove this

src/api/pastel-rpc/sent-tx-store.ts
```diff
@@ -0,0 +1,109 @@
...
+    this.expiryheight = 0
+    this.txid = ''
+    this.walletconflicts = []
+    this.time = 0
+    this.timereceived = 0
+    this.vjoinsplit = []
+    this.size = 0
+    this.lastblock = ''
+  }
+}
```
br4in3x: This has to be refactored to be a type

src/api/pastel-rpc/sent-tx-store.ts
```diff
@@ -0,0 +1,109 @@
...
+      transction.detailedTxns[0].address = s.address
+      transction.detailedTxns[0].amount = s.amount
+      transction.detailedTxns[0].memo = s.memo
+      return transction
+    })
+  } catch (err) {
+    // If error for whatever reason (most likely, file not found), just return an empty array
+    return []
+  }
+}
```
br4in3x: This is not related to RPC. Can we move it to the feature that is using this function?

src/features/pastelDB/pastelDBThread.tsx
```diff
@@ -0,0 +1,227 @@
...
+  getWalletTransaction,
+  listAddresses,
+  listTransactions,
+  listUnspent,
+} from '../../api/pastel-rpc/network-stats/network-stats'
+import {
+  TRawMempool,
+  TSinceblockTransaction,
+} from '../../api/pastel-rpc/network-stats/type'
+import { RPCConfig } from '../../legacy/components/AppState'
```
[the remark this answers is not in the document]
br4in3x: @Dinisueldanov seems like this was forgotten

src/features/pastelDB/pastelDBThread.tsx
```diff
@@ -0,0 +1,227 @@
...
+}
+
+interface fetchFuncConfig {
+  pastelDB: Database
+  rpcConfig: RPCConfig
+}
+
+const fetchStatisticinfo = async (props: fetchFuncConfig) => {
+  const results = await getStatisticInfo(props.rpcConfig)
+  insertStatisticDataToDB(props.pastelDB, results.hashrate, results.difficulty)
```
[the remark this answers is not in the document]
br4in3x: @Dinisueldanov seems like this was forgotten

src/features/pastelDB/__tests__/pastelDBThread.test.ts
```diff
@@ -623,9 +623,7 @@ describe('PastelDBThread', () => {
       expect(fetchTotalbalanceSpy).toHaveBeenCalled()
       expect(fetchListaddressesSpy).toHaveBeenCalled()
     } catch (e) {
-      expect(e).toEqual(
-        'Wrong API use : tried to bind a value of an unknown type (undefined).',
-      )
+      expect(e)
```
br4in3x, line 626: This doesn't seem right. Here should be an assertion

br4in3x (commented):
One last thing

br4in3x (approved):

PR #48, opened 2021-05-01
Wallet self updating

ngvtuan:
##### Commits
- [x] Make wallet self-updating so it updates the same way Chrome and VScode do
- [x] Done for window version

##### FreedCamp
- [x] [Make wallet self-updating so it updates the same way Chrome and VScode do](https://freedcamp.com/view/2956183/tasks/38537599)


br4in3x:
Perhaps we can update this to be a "Draft PR" if this is still in progress?

forge.config.js
```diff
@@ -40,19 +43,32 @@ module.exports = {
...
+    //   identity: 'Developer ID Application: Felix Rieseberg (LT94ZKYDCJ)',
+    //   'hardened-runtime': true,
+    //   entitlements: 'static/entitlements.plist',
+    //   'entitlements-inherit': 'static/entitlements.plist',
+    //   'signature-flags': 'library',
+    // },
+    // osxNotarize: {
+    //   appleId: 'felix@felix.fun',
+    //   appleIdPassword: 'my-apple-id-password',
+    // },
```
br4in3x, line 56: Do we need this? If this is going to be needed in the future, let's leave it

forge.config.js
```diff
@@ -40,19 +43,32 @@ module.exports = {
...
+        return {
+          exe: `${package.name}.exe`,
+          setupIcon: './static/icons/icon.ico',
+          loadingGif: './static/icons/icon.gif',
+          iconUrl:
+            'https://raw.githubusercontent.com/pastelnetwork/pastel-electron-wallet/master/static/icons/icon.ico',
+          title: package.productName,
+          setupExe: `${package.productName} Setup - v${package.version}.exe`,
+          skipUpdateIcon: true,
+        }
```
br4in3x, line 71: Could you explain what's the reason for this change?

src/features/Common.module.css
```diff
@@ -0,0 +1,38 @@
+.highlight {
```
br4in3x, line 1: Let's put it in the common folder

src/features/updateToast/UpdateToast.tsx
```diff
@@ -0,0 +1,57 @@
...
+    shell.openExternal(`https://github.com/${pkg.repoName}/releases`)
+  }
+
+  const handleUpdate = () => {
+    dispatch(closeUpdateToast())
+    ipcRenderer.send('restart_app')
+  }
+
+  if (!opened) {
+    return <></>
```
br4in3x, line 25:
```suggestion
    return null
```

src/features/updateToast/UpdateToastSlice.ts
```diff
@@ -0,0 +1,27 @@
...
+    },
+    closeUpdateToast(state: IUpdateToastState) {
+      state.opened = false
+    },
+  },
+})
+
+export const updateToastReducer = updateToastSlice.reducer
+
+export const { openUpdateToast, closeUpdateToast } = updateToastSlice.actions
```
br4in3x, line 27: I think nothing else is going to use the `updateToast`, so it's okay to have it's state through `useState`

src/main.ts
```diff
@@ -169,3 +174,67 @@ app.on('activate', () => {
...
+    }/${app.getVersion()}`
+
+    autoUpdater.setFeedURL({
+      url: feedURL,
+      serverType: 'default',
+    })
+    autoUpdater.checkForUpdates()
+    setInterval(() => {
+      autoUpdater.checkForUpdates()
+    }, 4 * 60 * 60 * 1000)
```
br4in3x, line 191: nit: can we create a constant with the interval name, like 
```ts
const oneHour = 60 * 60
```

src/main.ts
```diff
@@ -169,3 +174,67 @@ app.on('activate', () => {
...
+
+autoUpdater.on('error', err => {
+  console.log('error', { err })
+})
+
+const checkSignatureExists = () => {
+  if (os.platform() === 'darwin') {
+    const certExists = fs.existsSync(
+      path.join(__dirname, 'static', 'entitlements.plist'),
+    )
```
br4in3x, line 234: This is not a certificate. This is an MacOS app permission manifest. Not sure if it's valid to use it for detecting signature

src/features/updateToast/UpdateToast.tsx
ngvtuan: Updated

src/main.ts
ngvtuan: Updated

src/main.ts
ngvtuan: Yeah, I think we don't need it. Removed

src/features/Common.module.css
ngvtuan: Updated

forge.config.js
ngvtuan: Yes we need it for MAC. and I'm waiting for the Signing Code

br4in3x (approved):

br4in3x (approved):

PR #52, opened 2021-05-04
Update About popup and load version from package.json

ngvtuan:
##### Commits
- [x] Update About popup, load version from package.json and load the year from getFullYear Method

##### FreedCamp
- [x] [Update About popup and load version from package.json](https://freedcamp.com/view/2956183/tasks/39655776)


br4in3x (changes requested):
Please rename everything to just "About" - no need to prefix with "pastel". So for example, `pastelAbout` becomes just `about`.

ngvtuan:
@br4in3x I have updated the file name. please check

br4in3x (approved):

PR #55, opened 2021-05-04
Add squoosh app

ngvtuan:
##### Commits
- [x] Add squoosh app

##### FreedCamp
- [x] [Add squoosh app](https://freedcamp.com/view/2956183/tasks/39603824)


br4in3x (dismissed):
I'm thinking we should combine all modals into one reducer, that will look like this:
```
{
  sqoosh: true,
  about: false,
  // ...
}
```
Where key is the modal name, and value is whether it's open or not.

The same will be for the modal component. We are starting to have too many separate modals. What do you think?

ngvtuan:
@br4in3x 

The same will be for the modal component. We are starting to have too many separate modals. What do you think? --> Yes we can improve the code follow that way and i will create new ticket to improve for the modals

src/main.ts
```diff
@@ -134,6 +136,19 @@ const createWindow = async () => {
...
+    })
+    // Create server
+    const server = http.createServer(function onRequest(req, res) {
+      serve(req, res, () => {
+        console.log('Create server')
+      })
+    })
+    // Listen
+    server.listen(5100)
+  })
```
br4in3x, line 151: I think the server is not necessary. You can request `file:///node_modules/squoosh/build/index.html` (or something like this) directly in the modal

package.json
```diff
@@ -66,8 +67,10 @@
     "redux-thunk": "2.3.0",
     "regenerator-runtime": "0.13.7",
     "request": "2.88.2",
+    "serve-static": "1.14.1",
     "sha256-file": "1.0.0",
     "sql.js": "pastelnetwork/sql.js#230351f",
+    "squoosh": "pastelnetwork/squoosh#dev",
```
br4in3x, line 73: Can we have a commit hash instead of the branch name, similar how it's done for `sql.js` please?

src/main.ts
ngvtuan: @br4in3x I have tried with this but it doesn't work, when you tried to add like that, I will be shown 404. Refer this https://github.com/GoogleChromeLabs/squoosh/issues/999

package.json
ngvtuan: I have updated to the hash

br4in3x (approved):

PR #56, opened 2021-05-04
Refactor expert console

westalert:
Refactor expert console


package.json
```diff
@@ -20,7 +20,7 @@
     "test": "jest",
     "format": "eslint --ignore-path=.gitignore --ext .ts,.tsx --fix . && prettier --ignore-path=.gitignore --loglevel=error --write .",
     "check-eslint": "eslint --ignore-path=.gitignore --ext .ts,.tsx --max-warnings=0 .",
-    "check-prettier": "prettier --ignore-path=.gitignore --list-different .",
+    "check-prettier": "prettier --ignore-path=.gitignore --list-different --write .",
```
br4in3x, line 23: Let's not change it please, this command only checks. If you need prettier to write to your files, you can use `format` command.

src/features/errorModal/ErrorModal.tsx
```diff
@@ -0,0 +1,53 @@
+import React from 'react'
```
br4in3x, line 1: Could we reuse `pastelModal` feature?

src/features/expertConsole/TerminalConsole.tsx
```diff
@@ -227,22 +241,31 @@ function TerminalConsole(props: TConsoleProps): JSX.Element {
...
       const text =
         typeof data === 'object'
           ? textAsTable(
               Array.isArray(data)
-                ? data.map(i => formatConsoleData(i))
-                : [formatConsoleData(data)],
+                ? data.length > 0
+                  ? data.map(i => formatConsoleData(i))
+                  : ['']
+                : [formatConsoleData(data || {})],
```
br4in3x, line 256: This is a bit complicated. Can we please break it down into multiple functions or statements?

src/features/expertConsole/rpc-services/api.ts
```diff
@@ -24,30 +23,41 @@ const getMessage = (statusCode: number, isECONNREFUSED: boolean) => {
...
+type dynamicAPI = {
+  [x in possibleMethods]: <T>(params: string[]) => Promise<T>
+}
+
+let apiMethods = {} as dynamicAPI
+
+METHODS.forEach((method: string) => {
+  apiMethods = {
+    ...apiMethods,
+    // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
```
br4in3x, line 38: Please do not disable eslint rules. If it complains, please fix/refactor the code to remove the complaint

src/features/pastelID/pastelIDSlice.ts
```diff
@@ -6,7 +6,7 @@ import {
   TRPCConfig,
```
br4in3x, line 6: Changes to PastelID are unrelated to this PR. Could you revert them?

src/legacy/containers/Root.tsx
```diff
@@ -1,15 +1,15 @@
 import React from 'react'
```
br4in3x, line 1: Could you explain why is this changing?

src/redux/store.ts
```diff
@@ -1,5 +1,6 @@
 import { Action, configureStore, ThunkAction } from '@reduxjs/toolkit'
```
br4in3x, line 1: We can use `pastelModal` I think instead of `errorModal`. In fact, `pastelModal` was refactored from `errorModal` to be generic modal.

PR #58, opened 2021-05-06
Add the full list of masternodes to the default pastel.conf so that it will immediately have 25+ known good peers to speed up the initial blockchain download process

ngvtuan:
##### Commits
- [x] Add the full list of masternodes to the default pastel.conf so that it will immediately have 25+ known good peers to speed up the initial blockchain download process

##### FreedCamp
- [x] [Add the full list of masternodes to the default pastel.conf so that it will immediately have 25+ known good peers to speed up the initial blockchain download process](https://freedcamp.com/view/2956183/tasks/38737870)


src/legacy/components/LoadingScreen.tsx
```diff
@@ -159,7 +160,7 @@ class LoadingScreen extends Component<any, any> {
     // Load the RPC config from pastel.conf file
     const pastelLocation = locatePastelConf()
     let confValues
-
+    await updateDefaultPastelConfig(pastelLocation)
```
br4in3x, line 163: We need to do try/catch here please. Don't forget that we have a specific way of logging the error: `console.warn(<class> <function> error: ${error.message})`

src/features/pastelConf/pastelConf.ts
```diff
@@ -0,0 +1,43 @@
+import fs from 'fs'
+
+import masternodes from './masternodes'
+
+export const updateDefaultPastelConfig = async (
+  pastelConfPath: string,
+): Promise<number> => {
```
br4in3x, line 7: Instead of a number, can you return something like `Success | Failure | EmptyConfig`

src/features/pastelConf/pastelConf.ts
```diff
@@ -0,0 +1,43 @@
...
+        if (masterNodesExist.indexOf(value) === -1) {
+          confContent += `${value}\n`
+        }
+      })
+    }
+
+    await fs.promises.writeFile(pastelConfPath, confContent)
+    return 1 // success
+  } catch (err) {
+    return 3
```
br4in3x, line 41: I think this can re-throw: `throw new Error('pastelConf updateDefaultPastelConfig error: ${error.message}')`

src/features/pastelConf/pastelConf.ts
```diff
@@ -0,0 +1,43 @@
+import fs from 'fs'
+
+import masternodes from './masternodes'
+
+export const updateDefaultPastelConfig = async (
+  pastelConfPath: string,
+): Promise<number> => {
+  if (!pastelConfPath) {
+    return 2 // pastelConfPath empty
+  }
```
br4in3x, line 10: Instead of doing this, I think we can try/catch `await fs.promises.readFile(pastelConfPath)`. I don't think someone will willingly call this function with an empty string

src/features/pastelConf/__tests__/pastelConf.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    let mResponse = 'erver=1\n'
+    masternodes.map(m => {
+      mResponse += `${m.label}=${m.value}\n`
+    })
+    jest.spyOn(fs.promises, 'readFile').mockResolvedValueOnce(mResponse)
+
+    const result = await updateDefaultPastelConfig('somePath')
+    expect(fs.promises.writeFile).toBeCalledWith('somePath', mResponse)
+
+    expect(result).toBe(1)
```
br4in3x, line 52: I think we need one more test when this function throws an error to assert the correct error message

src/features/pastelConf/pastelConf.ts
ngvtuan: I have  updated for this one

src/features/pastelConf/pastelConf.ts
ngvtuan: I have  updated for this one

src/features/pastelConf/__tests__/pastelConf.test.ts
ngvtuan: I have  updated for this one

src/legacy/components/LoadingScreen.tsx
ngvtuan: I have  updated for this one

src/features/pastelConf/pastelConf.ts
ngvtuan: I have  updated for this one

src/features/pastelConf/__tests__/pastelConf.test.ts
```diff
@@ -46,9 +45,17 @@ describe('pastelConf/updateDefaultPastelConfig', () => {
...
     jest.spyOn(fs.promises, 'readFile').mockResolvedValueOnce(mResponse)
 
-    const result = await updateDefaultPastelConfig('somePath')
+    await updateDefaultPastelConfig('somePath')
     expect(fs.promises.writeFile).toBeCalledWith('somePath', mResponse)
+  })
 
-    expect(result).toBe(1)
+  test('catches an error if pastel.conf is undefined', async () => {
+    try {
```
br4in3x, line 53: For tests that assert in catch block, we need `expect.hasAssertions` because we want this test to throw an error that will go into catch block. `expect.hasAssertions` will fail the test if an error is not thrown

ngvtuan: I have updated for this

br4in3x (approved):

PR #61, opened 2021-05-12
Add ability to open a link from a website

ngvtuan:
##### Commits
- [x] Add ability to open a link from a website (such as the Pastel Explorer) to a specific location in the wallet app
- [x] Add sample screens and update the code to open the special screen
- [x] Update Prop-types
- [x] Remove console and fix mac can't open the screen

##### FreedCamp
- [x] [Add ability to open a link from a website (such as the Pastel Explorer) to a specific location in the wallet app](https://freedcamp.com/view/2956183/tasks/39753564)


src/main.ts
```diff
@@ -169,3 +216,25 @@ app.on('activate', () => {
...
+      redirectDeepLinkingUrl(deeplinkingUrl.toString(), mainWindow)
+    }
+  })
+})
+
+ipcMain.on('app-ready', () => {
+  if (mainWindow && deeplinkingUrl) {
+    redirectDeepLinkingUrl(deeplinkingUrl.toString(), mainWindow)
+  }
+})
```
br4in3x, line 240: I think we need to start putting stuff from the main file to a different folder, because `main` becomes a very large and unstable blob of code. Can we move this, let's say, to the `features/deepLinking` ? I'm not sure how right is that, I'll let you decide where it should go, but IMO having it in features is okay.

Also, having these function separately will allow to test them as well

src/main.ts
```diff
@@ -14,8 +14,52 @@ import http from 'http'
...
+        mainWindow.restore()
+      }
+      mainWindow.focus()
+      redirectDeepLinkingUrl(deeplinkingUrl.toString(), mainWindow)
+    }
+  })
+} else {
+  app.quit()
+}
+
```
br4in3x, line 62: I think we need to start putting stuff from the main file to a different folder, because `main` becomes a very large and unstable blob of code. Can we move this, let's say, to the `features/deepLinking` ? I'm not sure how right is that, I'll let you decide where it should go, but IMO having it in features is okay.

Also, having these function separately will allow to test them as well

src/main.ts
br4in3x, line 240: Same as above. I think handlers for `ipcMain.on` and `app.on` events can go to the feature folder and be tested

src/main.ts
ngvtuan: I have updated the code structure and move to the deepLinking folder

src/main.ts
ngvtuan: Updated and created test

src/main.ts
```diff
@@ -111,14 +85,15 @@ const createWindow = async () => {
   w.loadURL(MAIN_WINDOW_WEBPACK_ENTRY)
 
   // Open the DevTools.
-  if (!app.isPackaged) {
-    w.webContents.openDevTools()
-  }
+  // if (!app.isPackaged) {
+  //   w.webContents.openDevTools()
+  // }
+  w.webContents.openDevTools()
```
br4in3x, line 91: This will open dev tools in the prod version of the wallet

src/main.ts
```diff
@@ -222,24 +197,69 @@ app.on('activate', () => {
...
+  },
+)
+
+autoUpdater.on('update-not-available', () => {
+  console.log('update-not-available')
+})
+
+autoUpdater.on('error', err => {
+  console.log('error', { err })
 })
```
br4in3x: Can we change this to `console.warn('autoUpdater error: ${err.message}', error)`

src/main.ts
```diff
@@ -222,24 +197,69 @@ app.on('activate', () => {
...
+      releaseNotes,
+      releaseName,
+      updateURL,
+    })
+  },
+)
+
+autoUpdater.on('update-not-available', () => {
+  console.log('update-not-available')
+})
```
br4in3x: Same as another comment about `console.log`. Please change it everywhere

src/features/deepLinking/mocks/browser-window.ts
```diff
@@ -0,0 +1,128 @@
+import { EventEmitter } from 'events'
+
+import { WebContentsMock } from './web-contents'
+
+export class BrowserWindowMock extends EventEmitter {
```
br4in3x, line 5: Btw, this should go to `<root>/__mocks__/electron.ts`, so it would be picked up by `jest` automatically. See how it's done here: https://github.com/pastelnetwork/pastel-electron-wallet/blob/393f8032c1c35678fa1ae502c962bde24c0bd551/__mocks__/electron.ts

src/features/deepLinking/mocks/web-contents.ts
```diff
@@ -0,0 +1,104 @@
+import { EventEmitter } from 'events'
+
+export class WebContentsMock extends EventEmitter {
+  public static fromId = jest.fn()
```
br4in3x, line 4: Same comment about `__mocks__` folder

src/main.ts
ngvtuan: I have removed this one to test for the build version. I have reverted it back. Sorry

src/main.ts
ngvtuan: Yes, i have updated

src/features/deepLinking/mocks/browser-window.ts
ngvtuan: Yes, i have updated

src/features/deepLinking/mocks/web-contents.ts
ngvtuan: Updated

src/main.ts
ngvtuan: Updated

src/main.ts
```diff
@@ -257,9 +256,9 @@ autoUpdater.on(
 )
 
 autoUpdater.on('update-not-available', () => {
-  console.log('update-not-available')
+  console.warn('update-not-available')
 })
 
 autoUpdater.on('error', err => {
-  console.log('error', { err })
+  console.warn('autoUpdater error: ${err.message}', err)
```
br4in3x, line 263: Could you please log this error everywhere in this PR? I see that like [#258](https://github.com/pastelnetwork/pastel-electron-wallet/pull/61/files/303efb93e167796023dc2d0c03f3aa199935a544..de08638d17d097333064a65af5824c089bfce818#diff-4fab5baaca5c14d2de62d8d2fceef376ddddcc8e9509d86cfa5643f51b89ce3dR258) is still wrong

src/features/deepLinking/__tests__/deepLinking.test.ts
```diff
@@ -1,5 +1,5 @@
+import { BrowserWindowMock } from '../../../../__mocks__/browser-window'
```
br4in3x, line 1: You don't need to import this any more even, jest will automock it for you. Just import it directly from electron. The import will be automatically modified to BrowserWindow from `__mocks__` directory

src/main.ts
ngvtuan: OK @br4in3x , I have removed the un-need console.log

src/features/deepLinking/__tests__/deepLinking.test.ts
ngvtuan: Updated

src/features/expertConsole/rpc-services/api.ts
```diff
@@ -55,11 +55,11 @@ METHODS.forEach((method: string) => {
           cancelToken: source.token,
         })
           .then(data => {
-            console.log('[RPC CALL SUCCESS] -', method, data?.data?.result)
+            console.warn('[RPC CALL SUCCESS] -', method, data?.data?.result)
             resolve(data?.data?.result)
           })
           .catch(error => {
-            console.log('[RPC CALL ERROR] - ', { ...error })
+            console.warn('[RPC CALL ERROR] - ', { ...error })
```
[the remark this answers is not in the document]
ngvtuan: I have reverted for this change

br4in3x (approved):

PR #67, opened 2021-05-17
Feat/loading UI

westalert:
add animation for loading UI.


src/legacy/Routes.tsx
```diff
@@ -336,9 +346,10 @@ class RouteApp extends React.Component<any, any> {
       info: newInfo,
     })
   }
-  setInfo = (newInfo: any) => {
+  setInfo = (newInfo: TWalletInfo) => {
     // If the price is not set in this object, copy it over from the current object
     const { info } = this.state
+    console.log('newInfo', newInfo)
```
adrianobetkal, line 352: Redundant console.log

src/legacy/Routes.tsx
```diff
@@ -610,8 +621,9 @@ class RouteApp extends React.Component<any, any> {
 
               <Route
                 path={routes.LOADING}
-                render={() => (
+                render={p => (
```
adrianobetkal, line 624: I don't know what `p` is, use more descriptive name

westalert: p: prop of route when rendering @adrianobetkal 
It's so easy for understand for developers.

src/legacy/Routes.tsx
westalert: okay, I will remove

src/legacy/Routes.tsx
adrianobetkal: The good practice is to never user letters or shorts. Any argument, const, let, etc names should be descriptive even if you think `It's so easy for understand for developers.`

adrianobetkal (approved):

PR #68, opened 2021-05-17
Onboarding eslint and login page

adrianobetkal:
Adjusted
- `eslint` and `prettier` configs
- `Input` component
- `Button` component
- `Theme`
- `Routes`

Added
- `Login` page


src/components/Input/Input.styles.ts
```diff
@@ -1,10 +1,18 @@
...
+export const InputContainer = styled.div`
+  position: relative;
+  display: flex;
+  justify-content: space-between;
+`;
+
+export const Input = styled.input<{ bordercolor: string }>`
   background-color: ${({ theme }) => theme.colors.white};
-  border: 1px solid #8e98a333;
+  border: 1px solid ${({ bordercolor }) => bordercolor};
```
tomek-ifm, line 13: Why not borderColor ?

adrianobetkal: Not sure, you're right, changed

PR #73, opened 2021-05-19
New Design Components and Theme

adrianobetkal:
Added
- `styled-component` library
- `icons` and `images`
- `Button` component with styles
- `Input` component with styles
- `Typography` component with styles
- `Theme` colors, shadows and typography

Adjusted
- `Features` import styles


package.json
```diff
@@ -74,6 +74,7 @@
     "sha256-file": "1.0.0",
     "sql.js": "pastelnetwork/sql.js#230351f",
     "squoosh": "pastelnetwork/squoosh#d44b1b4",
+    "styled-components": "^5.3.0",
```
br4in3x, line 77: Could you please specify the exact version of the package? This would prevent unwanted updates, and this is the pattern we're following

br4in3x (commented):
Looks good to me! Could we please move the `common` folder out of features, because it's not a feature, but shared code?

package.json
adrianobetkal: Sure, changed.

PR #75, opened 2021-05-20
Add to Misc Tools menu menubar the Glitch library

ngvtuan:
##### Commits
- [x] Add to Misc Tools menu menubar the Glitch library

##### FreedCamp
- [x] [Add to Misc Tools menu menubar the Glitch library](https://freedcamp.com/view/2956183/tasks/39163963)


src/features/serveStatic/ServeStatic.tsx
```diff
@@ -0,0 +1,55 @@
...
+      process.resourcesPath,
+      '/app.asar/.webpack/renderer/static/squoosh',
+    )
+    glitchStaticPath = path.join(
+      process.resourcesPath,
+      '/app.asar/.webpack/renderer/static/glitch',
+    )
+  }
+  setupServeStatic(squooshStaticPath, 5200)
+  setupServeStatic(glitchStaticPath, 5300)
```
br4in3x, line 21: I think we can have the port as a constant in the feature folder of every tool, and call it here like `sqoosh.staticPort`, `glicth.staticPort`

src/features/loading/LoadingScreen.tsx
```diff
@@ -120,8 +121,8 @@ class LoadingScreen extends Component<TLoadingProps, TLoadingState> {
 
   loadingConfigs = async () => {
     const success = await this.ensurePastelParams()
-
     if (success) {
+      initServeStatic(remote.app.isPackaged)
```
br4in3x, line 125: I think you can call ipcMain here and dispatch "loadConfigSuccess" event, and in the `main.ts`, listen for that event and call `initServeStatic(remote.app.isPackaged)`. It would be cleaner. Read more about ipcMain here: https://www.electronjs.org/docs/api/ipc-main

src/features/loading/LoadingScreen.tsx
```diff
@@ -290,11 +291,12 @@ class LoadingScreen extends Component<TLoadingProps, TLoadingState> {
         // And reply that we're all done.
         ipcRenderer.send('appquitdone')
       }
+
+      closeServeStatic()
```
br4in3x, line 295: Similar to the comment above, we should call this in `main.ts` on the event when the app shutting down

src/features/serveStatic/ServeStatic.tsx
ngvtuan: It has been updated

src/features/loading/LoadingScreen.tsx
ngvtuan: It has been updated

src/features/loading/LoadingScreen.tsx
ngvtuan: It has been updated

br4in3x (approved):

westalert (approved):

PR #79, opened 2021-05-24
Create Message Tab 

westalert:
Create messages tab 
 - can send messages
 - can upload voice.
 - can upload image
 - can use emoji


src/features/messages/MessagesStore.ts
```diff
@@ -0,0 +1,153 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit'
```
br4in3x, line 1: Can we please have tests for this? An example can be found in https://github.com/pastelnetwork/pastel-electron-wallet/tree/a69d01a4d68460eda71bde7866fc81f9483127ae/src/features/pastelPrice/__tests__

src/features/messages/components/Compose/index.tsx
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+  const onChangeShowImage = () => {
+    const element = document.getElementById('input-compose-image')
+    element?.click()
+  }
+
+  const onChangeImageCompose = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const files = e.target.files
+    if (files) {
```
br4in3x, line 100:
```suggestion
    if (files.length > 0) {
```

src/features/messages/components/Compose/index.tsx
```diff
@@ -0,0 +1,259 @@
...
+      Array.from(files).forEach((file: File) => {
+        const ext = file.name.split('.').pop()?.toLowerCase() || ''
+        if (!['jpg', 'png', 'jpeg'].includes(ext)) {
+          console.log('File invalid')
```
br4in3x, line 104: We should do something better than a `console.log`. Perhaps show an error to a user?

src/features/messages/components/Compose/index.tsx
```diff
@@ -0,0 +1,259 @@
...
+    setAudio(recordedBlob.blobURL)
+  }
+
+  const startRecording = () => {
+    setRecord(true)
+  }
+
+  const stopRecording = () => {
+    setRecord(false)
+  }
```
br4in3x, line 140: Do we really need these functions? They only wrap a single call to the `set...` methods

package.json
```diff
@@ -48,20 +48,24 @@
...
     "hex-string": "1.0.3",
     "history": "4.10.1",
     "http": "0.0.1-security",
+    "immer": "9.0.2",
     "ini": "2.0.0",
     "javascript-terminal": "1.1.1",
     "js-base64": "3.5.2",
     "js-sha256": "0.9.0",
     "libsodium-wrappers-sumo": "0.7.6",
+    "moment": "2.29.1",
```
br4in3x, line 61: We already have `[day.js](https://day.js.org/)`, I think it's pretty much a replacement for `moment`

package.json
```diff
@@ -48,20 +48,24 @@
     "electron-log": "4.3.2",
     "electron-squirrel-startup": "1.0.0",
     "electron-store": "5.1.1",
+    "emoji-mart": "^3.0.1",
```
br4in3x, line 51: Could we remove this from emoji selector?
image

br4in3x: Emoji selector doesn't disappear when I click something else. Same for audio selector

br4in3x: Also, audio selector doesn't work for me - when I press the "mic" icon, the recording doesn't start.

br4in3x (commented):
Could you please fix this warning?
image

package.json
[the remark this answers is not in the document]
westalert: @berkcent okay. I will fix.

src/features/messages/MessagesStore.ts
westalert: Got it.

ngvtuan (commented):
Some emoji icons don't show after I pressed enter
icon-showing-wrong

src/features/messages/Messages.tsx
```diff
@@ -0,0 +1,108 @@
+import React, { useEffect } from 'react'
+import { connect } from 'react-redux'
+
+import { RootState } from '../../redux/store'
+import ConversationList from './components/ConversationList'
+import MessageList from './components/MessageList'
+import Skeleton from './components/skeletons'
+// import style from './Messages.module.css'
```
br4in3x, line 8: Please remove all commented code that is not needed

src/features/messages/conversation-mock.ts
```diff
@@ -0,0 +1,18 @@
+import { TConversation } from './MessagesStore'
+
+interface RESPONSE {
```
br4in3x, line 3: This shouldn't be capitalized

br4in3x (commented):
Looks good to me, except that audio recording still works only every other time, and image upload doesn't seem to work

PR #84, opened 2021-05-25
Header component

sanjanadhikari1:
- [x] install tailwindcss on the project using Cris's code.
- [ ] create pixel perfect header component
- [ ] make it responsive using tailwindcss
- [ ] Fix 2 issues mentioned.

**FreedCamp**
[https://freedcamp.com/view/2956183/tasks/40068433]()


CrazyOptimist (changes requested):
I failed to resolve conflicts on tailwind configuration, and I think it's better to be done on your side, please check the conflicts and adjust your code according to the existing configuration in the `new_development`

PR #86, opened 2021-05-25
Onboarding

adrianobetkal:
Added

- Checkbox 
- FormLoading
- Radio
- Tooltip
- Button
- Input
- Typography

- cryptoKeys
- login
- newPassword
- passwordRecovery
- register with 5 steps

1. - backupSteps
2. - feeSteps
3. - paymentSteps
4. - promotionSteps
5. - registerSteps

- registrationPending
- registrationSuccessful
- welcome


Dicklesworthstone (dismissed):

src/common/routes/Routes.tsx
```diff
@@ -0,0 +1,72 @@
+/* eslint-disable */
```
br4in3x, line 1: We shouldn't have this

adrianobetkal: I agree, but this is only for the line below with `any`, so will be removed after redux flow implementation
`any // remove this any after user login is taken by redux`

adrianobetkal: I agree, but this is only for the line below with `any`, so will be removed after redux flow implementation
`any // remove this any after user login is taken by redux`

br4in3x: In this case, could we use `// eslint-disable-next-line @typescript-eslint/no-explicit-any`? That would disable the rule only for the line where it should be ingored

adrianobetkal: Sure, I just pushed changes. Unfortunately, merging master failed, because of other files. Let me know if you know why it happened.

Screenshot 2021-05-26 at 07 59 11

tomek-ifm (approved):

sosnomIntent (approved):

PR #100, opened 2021-05-27
complete autocomplete component

sanjanadhikari1:
- [ ] Developed AutoComplete Components using TailwindCss.
image
- [ ] Fixed some tailwindcss config file for resolving conflicts with others
image
Make those pixels
- [ ] Fixed some tailwind css in Header Component according to tailwind config file.

**FreedCamp**
[https://freedcamp.com/view/2956183/tasks/40128326](url)


dbizhev (commented):
In my experience, normally using pixels is not such a good idea for browsers, because it’s a fixed unit and does not take into consideration browser zoom. As far as I know, Electron is kind of a browser window, so I think it gets back to which approach we want to take, it’s a long discussion like here https://stackoverflow.com/questions/11799236/should-i-use-px-or-rem-value-units-in-my-css.

My opinion is that tailwind already uses rem for every value they have, and I also believe it to be good practice to use rem. If the user has a big monitor and zoom into the wallet, using rem guarantees that everything stays quite the same because rem is not fixed like pixels

CrazyOptimist:
> * [ ]  Fixed some tailwindcss config file for resolving conflicts with others
>   image
>   Make those pixels

@sanjanadhikari1 In my opinion, this is the right approach for pixel perfect design.
We need more reviewers to discuss maybe?
Please read [here](https://tailwindcss.com/docs/customizing-spacing) for more information.

CrazyOptimist (approved):
Looks good to me. 👍

reacteer (approved):

PR #105, opened 2021-05-28
Fix/new development issues 40138259

dbizhev:
- [x] Can we fix the build warning? please see `Build warning with tailwind.png`

- [x] Can we remove the un-use code in the config? please see `remove un-use code.jpg`
remove un-use code.jp

- [x] Why do we need to use both `rem` and `px`? Can we only use one? please see `inconsistency unit.jpg`

- [x] Can we use `xxl` instead of `2xl`? It does not make sense here because 2xl should be `2880px`? please see `xxl instead of 2xl.jpg`

- [x] Can we update to `14, 15, 16` to make sure it consistent with other sections? please see `14 15 16 to make it consistency with others.jpg`

- [x] It has a lot of duplicate color in the `tailwind.config.js` file, Can we update this? please see `duplicate color.jpg`

- [x] We need to clear up the `index.css` file, For now, we are using CSS and import tailwind to that file, I don't think this is the right way? please see `duplicate color.jpg`

- [x] As we have agreed before each component should has an `index.ts` file. please see `missing index.ts file.jpg`

- [x] C an we add a prefix for our CSS class? It will be help full to identify our branding. I think `psl` is a good option https://tailwindcss.com/docs/configuration#prefix


CrazyOptimist:
Fixes are looking good to me, all resolved.
Let's convert the `px` to `rem` in the tailwind config now.

bulatsalyakhovdev (approved):

CrazyOptimist (approved):
Looks good to me!

src/common/components/Header/HeaderScreen.tsx
```diff
@@ -44,7 +44,7 @@ const SidebarMenuItem = ({
       className={[
         styles.headermenuitem,
         activeColorClass,
-        'mr-8px md:mr-13px lg:mr-28px',
+        'psl-mr-8px md:psl-mr-13px lg:psl-mr-28px',
```
ngvtuan, line 47: Remove alls `psl` prefix, please. As we have discussed in the group

src/legacy/Routes.tsx
```diff
@@ -109,11 +109,11 @@ class RouteApp extends React.Component<any, any> {
...
-    this.rpcRefreshIntervalId = window.setInterval(() => {
-      if (this.state.rpcConfig.username) {
-        rpc.refresh()
-      }
-    }, 10000)
+    // this.rpcRefreshIntervalId = window.setInterval(() => {
+    //   if (this.state.rpcConfig.username) {
+    //     rpc.refresh()
+    //   }
+    // }, 10000)
```
ngvtuan, line 116: why did we need to remove this? Is this remove for the development mode? If yes, can we use `if (!app.isPackaged) {` ?

please refer to: https://github.com/pastelnetwork/pastel-electron-wallet/blob/master/src/main.ts Line 87

tailwind.config.js
```diff
@@ -1,59 +1,72 @@
...
+        '68px': '4.25rem',
+        '69px': '4.313rem',
+        '70px': '4.375rem',
+        '82px': '5.125rem',
+        '111px': '6.938rem',
+        '220px': '13.75rem',
+        '300px': '18.75rem',
+        '335px': '20.938rem',
+        '352px': '22rem',
+        30: '7.5rem',
```
ngvtuan, line 47: can we change this format like others? 30 is not make sense, I think it should be `[number]px`

tailwind.config.js
```diff
@@ -1,59 +1,72 @@
...
-  },
-  purge: [],
+  purge: [
+    './src/**/*.ts',
+    './src/**/*.tsx',
+    './src/**/*.js',
+    './src/**/*.jsx',
+    './src/**/*.html',
+  ],
+  prefix: 'psl-',
```
ngvtuan, line 9: Please remove the prefix. As we have discussed in the group

tailwind.config.js
```diff
@@ -75,14 +90,52 @@ module.exports = {
...
+        h1: '2rem',
+        h2: '1.5rem',
+        h3: '1.375',
+        h4: '1.125',
+        h5: '1rem',
+        h6: '0.875rem',
+        15: '0.938rem',
+      },
+      borderRadius: {
+        '4px': '4px',
```
ngvtuan, line 137: Can we update to `rem` ?

CrazyOptimist:
Please make sure that you name the similar branches `bugfix/abc-xyz`, instead of `fix/abc-xyz` from the next time.  
We have this convention, in case you missed it.  

```
# for new features
`feature/meaningful-name-freedcamp_task_number`
# for bugfixes
`bugfix/bug-name-freedcamp_task_number`
# for experimental changes
`experiment/meaningful-name-freedcamp_task_number`
# for chores
`chore/chore-name-freedcamp_task_number`
# for releases
`release/0.1.x`
# all the commit messages should start with lower case letter 
```

PR #110, opened 2021-05-31
Update the version number from 1.2.0 to 1.2.1

ngvtuan:
##### Commits
- [x] Update the version number from 1.2.0 to 1.2.1

##### FreedCamp
- [x] [Release a new version of the pastel wallet app](https://freedcamp.com/view/2956183/tasks/40208448)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "pastelwallet",
   "productName": "Pastel Wallet Fullnode",
-  "version": "1.2.0",
+  "version": "1.2.1",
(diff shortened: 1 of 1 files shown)


br4in3x (approved):

CrazyOptimist:
I think it's better if we do this right after releasing, and it would be done by the person who published the release. Without another PR I mean.
Please refer to [here](https://github.com/slack-ruby/slack-ruby-bot-server-events/blob/master/RELEASING.md).

CrazyOptimist (approved):
LGTM

PR #115, opened 2021-06-01
fix when the user has installed the previous version of the wallet

ngvtuan:
##### Commits
- [x] fix when the user has installed the previous version of the wallet

##### FreedCamp
- [x] [Test and fix when the user has installed the previous version of the wallet](https://freedcamp.com/view/2956183/tasks/40216855)


prodev1101 (approved):
working as expected

muhadovic (approved):

PR #125, opened 2021-06-04
Feature/transactions in block chart

Dinisueldanov:
### Summary
- added transactions in block chart.
- added tests.

### Screenshots
txinblock


reacteer:
image

I think it will be great if we can see entire graph adjusting y axis.

ngvtuan (commented):
transactions-in-block
Please remove all un-need code in this PR. 

transactions-in-block-dub
Can we remove the `Transaction` active state? 

change_file_name
Can we change the file name to `PSL_transactions_in_block_...`? 

set_default
Can we set `2h` as default? Because other charts are set the `first left option` as a default. So I think we can make this update to consistent with other charts.

ngvtuan (approved):

westalert (approved):

PR #129, opened 2021-06-07
Feature/transactions per second chart

Dinisueldanov:
### Summary
- added transactions per second chart.
- added tests.

### Screenshot
transactionspersecond


src/features/pastelDB/__tests__/pastelDBLib.test.ts
```diff
@@ -1354,4 +1354,50 @@ describe('managePastelDatabase', () => {
       )
     }
   })
+
+  test('getTransactionsDataFromDBByPeriod should works correctly', async () => {
```
ngvtuan, line 1358: I think  we  need to create 2 test cases  for  this  test
- Correct case  
- Error case

ngvtuan (approved):

sanjanadhikari1 (approved):

PR #130, opened 2021-06-07
Feature/table component

sanjanadhikari1:
- [ ] complete table component and add it to the storybook
- [ ] complete checkbox component and add it to the storybook
- [ ] complete tooltip component and add it to the storybook


ihor-stack (commented):
There is conflict on this PR.

ihor-stack (approved):

src/common/components/Table/Table.stories.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+import Table, { TTableProps } from './index'
+
+export default {
+  title: 'Table',
+  component: Table,
+} as Meta
+
+const Template: Story<TTableProps> = ({ ...args }: TTableProps) => {
+  //   const [checked, setChecked] = useState(isChecked)
```
CrazyOptimist, line 12: could you remove unused code.

CrazyOptimist (changes requested):
image
* I think the caret should be aligned at the middle vertically.
* For table story, could we use string data for one column instead of both numbers?

* Tooltip is not fully appearing in the storybook, could you put it a bit down? It's the same in the doc tab.

src/common/components/Table/Table.stories.tsx
sanjanadhikari1: fixed

CrazyOptimist (approved):

PR #132, opened 2021-06-07
roll back to use legacy routes

CrazyOptimist:
There are routing files, which we needed to rollback to resolve comments from @ngvtuan and we won't modify legacy folder anymore.


sanjanadhikari1 (approved):
It looks good for me

reacteer (approved):

PR #133, opened 2021-06-08
fix 5 issues on ticket.

sanjanadhikari1:
- [ ] Background color and text color doesn't the same
- [ ] Missing underline and font size is not the same as the design 
- [ ] Position is not right and we must use SVG for the icon 
- [ ] The red icon too small and not the same as the design
- [ ] Position is not the same as the design 

**FreedCamp**
[https://freedcamp.com/view/2956183/tasks/40138661](url)


reacteer:
image

It will be great if we can have pointer when hovering these buttons.

reacteer (approved):

ngvtuan (approved):

PR #142, opened 2021-06-13
2.3 My Profile pages

reacteer:
My profile pages - 2.3_my profile, 2.3_my profile_empty, 2.3.1_my profile_edit in figma. 
Note: there might be some mis-matching with figma sheet due to the design changes as known.

https://freedcamp.com/view/2956183/tasks/40112264

image
image
image


src/features/profile/myProfile/MyProfile.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  const onTabToggle = (index: number) => {
+    setTab(index)
+  }
+
+  return (
+    <div className='mx-auto w-full bg-gray-f8 text-gray-23'>
+      <div className='wrapper pt-0 h-35px flex items-center'>
+        <div className='text-sm text-gray-71'>My account / General</div>
+      </div>
```
ArthurVogelsang, line 20: @reacteer I think this is static. Please use Breadcrumbs component, refer PR#144

ArthurVogelsang (changes requested):

ArthurVogelsang:
@reacteer I think your page should only show when `general` tab is active, if `security` is active, should show the security page. Please check this on `new_development` branch
image

ArthurVogelsang:
@reacteer I think this responsiveness is not good, I think should keep the design layout til about 1200px.
image
I think this should have `max-height`, if so, not drop down too low like now

image

src/legacy/assets/img/camera.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="14" height="13" viewBox="0 0 14 13" fill="none" xmlns="http://www.w3.org/2000/svg">
```
CrazyOptimist, line 1: Why to place icons to legacy?

tailwind.config.js
```diff
@@ -255,6 +263,8 @@ module.exports = {
         '15px': '0.938rem',
         '26px': '1.625rem',
         '32px': '2rem',
+        12: '0.75rem',
+        32: '2rem',
```
CrazyOptimist, line 267: Please update code to use existing sizes, so as '32px' already exists, no need for the same without "px".
As other sizes have "px", rename 12 to 12px as well

CrazyOptimist (changes requested):

ArthurVogelsang:
@reacteer Please fix the conflicts, and take a look portfolio page deeply

src/legacy/assets/img/camera.svg
reacteer: @CrazyOptimist , I think this came from other's work.

CrazyOptimist: Please copy it into our new icon directory, your work will broken later otherwise, because legacy folder will be removed after all.

src/features/profile/components/Categories.tsx
```diff
@@ -0,0 +1,76 @@
...
+      )}
+      {isAdding && (
+        <div className='flex'>
+          <div className='px-2 mx-2 rounded-full text-gray-dd border border-gray-dd flex text-sm items-center  mb-2'>
+            <div
+              onKeyPress={onAddCategory}
+              ref={inputRef}
+              className='h-6 text-gray-4a min-w-32px outline-none'
+              contentEditable={true}
+            />
```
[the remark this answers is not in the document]
reacteer: @CrazyOptimist , If I use input here, it can't be resizable according to its content and it will look ugly.

src/legacy/assets/img/camera.svg
CrazyOptimist: Also the icon would be good to be added to storybook

ArthurVogelsang:
@reacteer I just fixed this issue
image

ArthurVogelsang (approved):

src/features/profile/components/Categories.tsx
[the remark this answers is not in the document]
CrazyOptimist: contentEditable is only for the case when we want to store html user input, for rich text editor for example. As with contentEditable user can submit raw html, it requires special treatment - whitelisting of allowed tags, removing styles, etc.

We have existing dependency "downshift" for creating customizable select inputs:
https://codesandbox.io/s/o4yp9vmm8z

Style and behavior of this library is fully customizable. It may be complicated, so let me know if you would need any help

muhadovic (changes requested):

CrazyOptimist (approved):
I see security issue fixed, good work. 👍

PR #145, opened 2021-06-14
2.4 screen 40112289

muhadovic:
https://freedcamp.com/dashboard/home/panel/task/40112289

image


src/common/components/Inputs/Input.tsx
```diff
@@ -40,14 +37,9 @@ const Input: React.FC<TInput> = React.forwardRef(
...
-    ref: React.LegacyRef<HTMLInputElement>,
+    ref,
   ) => {
-    const classes = cn(
-      {
-        'relative flex items-center': true,
-      },
-      className,
-    )
+    const classes = cn({ 'relative flex items-center': true }, className)
```
CrazyOptimist, line 42: nit: `cn` can accept classes without object syntax for simplicity:

```
const classes = cn('relative flex items-center', className)
```

src/common/components/Inputs/InputNumberFormat.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+import Input, { TInput } from './Input'
+import NumberFormat from 'react-number-format'
+
+type Override<T1, T2> = Omit<T1, keyof T2> & T2
```
CrazyOptimist, line 5: nit: nice helper! I think it may be useful for other components as well, would be good to have a file with such helpers, in common/types.ts for example

src/common/components/Inputs/Input.tsx
```diff
@@ -1,16 +1,14 @@
...
 import TimesIcon from '../../assets/icons/ico-times.svg'
-import NumberFormat from 'react-number-format'
 
 export type TInput = {
   className?: string
-  type?: 'text' | 'password' | 'number'
-  kind?: 'numberFormat'
-  prepend?: ReactSVGElement
-  append?: ReactSVGElement
+  type?: string
```
CrazyOptimist, line 9: string can be accept any value, by default it is "text" in this component, but I see that InputPassword component provides "string" as a type.

Please turn it to the union of acceptable values, so typescript could check for valid types

src/common/components/Inputs/InputPassword.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react'
+import Input, { TInput } from './Input'
+import EyeIcon from '../../assets/icons/ico-eye.svg'
+import Icon from '../Icon'
+
+const InputPassword: React.FC<TInput> = props => {
+  const [type, setType] = React.useState<string>('password')
+
+  const toggleType = (): void =>
+    setType(type => (type === 'password' ? 'string' : 'password'))
```
CrazyOptimist, line 10: type string seems to be incorrect

CrazyOptimist (approved):

src/common/components/Inputs/InputNumberFormat.tsx
muhadovic: I added types.ts in common/utils

Dicklesworthstone (approved):

PR #146, opened 2021-06-14
Api pastel rpc no response error

ngvtuan:
##### Commits
- [x] Fix the issue "api/pastel-rpc no response error"

##### FreedCamp
- [x] [Fix the issue "api/pastel-rpc no response error"](https://freedcamp.com/view/2956183/tasks/40483206)

### DIFF
--- a/src/api/pastel-rpc/rpc.ts
+++ b/src/api/pastel-rpc/rpc.ts
@@ -35,7 +35,7 @@ export async function rpc<T>(
           err.response?.data,
         )}. Status code: ${JSON.stringify(err.response?.status)}`,
       )
-      throw new Error(`api/pastel-rpc server error: ${err.response?.message}`)
+      throw new Error(`api/pastel-rpc server error: ${err.message}`)
     }
 
     if (err.request) {
@@ -45,9 +45,7 @@ export async function rpc<T>(
           err.request,
         )}.`,
       )
-      throw new Error(
-        `api/pastel-rpc no response error: ${err.request?.message}`,
-      )
+      throw new Error(`api/pastel-rpc no response error: ${err.request}`)
     }
 
     log.error(
--- a/src/api/pastel-rpc/rpc.ts
+++ b/src/api/pastel-rpc/rpc.ts
@@ -31,21 +31,17 @@ export async function rpc<T>(
   } catch (err) {
     if (err.response) {
       log.error(
-        `api/pastel-rpc server error: ${
-          err.response?.message
-        }. Response: ${JSON.stringify(err.response)}. Detail: ${JSON.stringify(
+        `api/pastel-rpc server error. Response: ${JSON.stringify(err.response)}. Error: ${JSON.stringify(
           err,
         )}`,
       )
-      throw new Error(`api/pastel-rpc server error: ${err.response.message}`)
+      throw new Error(`api/pastel-rpc server error: ${err.response?.message}`)
     }
 
     if (err.request) {
       // The request was made but no response was received
       log.error(
-        `api/pastel-rpc no response error: ${
-          err.request?.message
-        }. Request: ${JSON.stringify(err.request)}. Detail: ${JSON.stringify(
+        `api/pastel-rpc no response error. Request: ${JSON.stringify(err.request)}. Error: ${JSON.stringify(
           err,
         )}`,
       )
@@ -55,7 +51,7 @@ export async function rpc<T>(
     }
 
     log.error(
-      `api/pastel-rpc error: Cannot connect to Pasteld. Detail: ${JSON.stringify(
+      `api/pastel-rpc error: Cannot connect to Pasteld. Error: ${JSON.stringify(
         err,
       )}`,
     )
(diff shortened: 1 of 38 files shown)


src/features/pastelID/PastelID.tsx
```diff
@@ -15,6 +19,18 @@ import { openPastelModal } from '../pastelModal'
 import styles from './PastelID.module.css'
 import { createPastelID, fetchPastelIDs } from './pastelIDSlice'
 
+const locatePastelConfDir = () => {
+  if (os.platform() === 'darwin') {
+    return path.join(remote.app.getPath('appData'), 'Pastel')
```
CrazyOptimist, line 24: Electron doc says that remote is deprecated, has performance and security concerns: https://electronjs.org/docs/api/remote

Can we use suggested ipc instead?

src/features/pastelID/PastelID.tsx
```diff
@@ -25,6 +41,16 @@ function passphraseStatusColor(validation: TPasswordStrengthResult) {
   return colors[validation.id]
 }
 
+function createPastelKeysFolder() {
+  const rootPath = locatePastelConfDir()
+  if (rootPath) {
+    const pastelKeysFolder = path.join(rootPath, 'pastelkeys')
+    if (!fs.existsSync(pastelKeysFolder)) {
+      fs.mkdirSync(pastelKeysFolder)
```
CrazyOptimist, line 49: Please use unblocking promise methods: fs.promises.access and fs.promises.mkdir instead. Sync methods are blocking main (UI) thread.

src/features/receive/Receive.tsx
```diff
@@ -59,6 +59,7 @@ export const Receive = (props: IReceiveProps): JSX.Element => {
     transactions,
   } = props
 
+  const [currentTab, setCurrentTab] = useState(0)
```
CrazyOptimist, line 62: Consider using typescript enum instead of a number. Because when I see `currentTab === 0 ` it's harder to see what is this for, better to have `currentTab === Tab.shielded`. Or ETab for enum since you prefer T for types.

src/renderer.tsx
```diff
@@ -34,6 +34,7 @@ import React from 'react'
 import { render } from 'react-dom'
 import { hot } from 'react-hot-loader' // has to stay first
 import { Provider } from 'react-redux'
+import log from 'electron-log'
```
romeerez, line 37: We at javascript are get used to console.log, without imports.

I think it would be better to assign log to console:
console.log = log.log;

CrazyOptimist, line 37: I suggest to assign needed log to console:
```
console.log = log.log
console.error = log.error
```
Because in javascript we get used to console.log, it's faster to use it rather then importing log

src/renderer.tsx
```diff
@@ -63,6 +64,8 @@ try {
   console.error(`PastelDB.getDatabaseInstance error: ${error.message}`)
 }
 
+log.transports.console.level = false
```
CrazyOptimist, line 67: Let's have console level = false only in production mode, to be able to see the log during development

And file.level = false for not in production, as we can see log in console during development and no need for log files

CrazyOptimist (changes requested):

src/features/pastelID/PastelID.tsx
ngvtuan: Fixed

src/features/pastelID/PastelID.tsx
ngvtuan: Fixed

src/features/receive/Receive.tsx
ngvtuan: Fixed

src/renderer.tsx
ngvtuan: Fixed

src/renderer.tsx
ngvtuan: Fixed

src/api/pastel-rpc/rpc.ts
```diff
@@ -29,15 +30,30 @@ export async function rpc<T>(
     })
   } catch (err) {
     if (err.response) {
+      log.error(
```
CrazyOptimist, line 33: please log.error to console.error as now electron log is assigned to a `console`, and in other places in this file as well

src/renderer.tsx
```diff
@@ -63,6 +66,24 @@ try {
   console.error(`PastelDB.getDatabaseInstance error: ${error.message}`)
 }
 
+ipcRenderer.on(
```
CrazyOptimist, line 69: This looks like a lot of code to just get current environment.
Is there a reason why we can't use process.env.NODE_ENV for simplicity?

ngvtuan: because it already has a property so that's why we have used it to consistent https://www.electronjs.org/docs/api/app#appispackaged-readonly

src/api/pastel-rpc/rpc.ts
ngvtuan: Fixed

CrazyOptimist (approved):

Dicklesworthstone (approved):

PR #150, opened 2021-06-16
Feature/nft/modals62 68

reacteer:
For testing, use `routes.PORTFOLIO_DETAIL` to see portfolio detail page.

And here are temporary buttons for testing modals.

image


src/features/nft/nftModals/AuthorshipClaimModal.tsx
```diff
@@ -0,0 +1,65 @@
...
+  handleClose,
+}) => {
+  const data = {
+    nftName: 'i.e banksy168',
+    pastelId: 'i.e banksy168',
+  }
+  const [text, setText] = useState('')
+
+  return (
+    <Modal
```
CrazyOptimist, line 21: just for future, consider using common/components/AnimatedModal, I think modal looks better with transition

CrazyOptimist (approved):

src/features/nft/nftModals/AuthorshipClaimModal.tsx
CrazyOptimist: Also with current modal the content will always be rendered, even though all modals are closed react will have to spend time to render them

sanjanadhikari1 (approved):

PR #157, opened 2021-06-18
disable console log in production

ngvtuan:
##### Commits
- [x] disable console log in production
- [x] Remove `^` in the package.json


reacteer (approved):

package.json
```diff
@@ -38,7 +38,7 @@
     "@hot-loader/react-dom": "17.0.1",
     "@react-pdf/renderer": "2.0.0",
     "@reduxjs/toolkit": "1.5.1",
-    "@tailwindcss/postcss7-compat": "^2.1.2",
+    "@tailwindcss/postcss7-compat": "2.1.2",
```
CrazyOptimist, line 41: It could be resolved when we merge `new_development` into master.
As long as you changed it in this PR, all the tailwind related deps should be updated to avoid confusion when we merge new design into master, you can check it in `new_development`.
Please make sure that you remove this dep, it's not needed anymore because we use `postcss8`

package.json
```diff
@@ -93,7 +93,7 @@
     "sql.js": "pastelnetwork/sql.js#230351f",
     "squoosh": "pastelnetwork/squoosh#b9388e1",
     "styled-components": "5.3.0",
-    "tailwindcss": "npm:@tailwindcss/postcss7-compat@^2.0.4",
+    "tailwindcss": "npm:@tailwindcss/postcss7-compat@2.0.4",
```
CrazyOptimist, line 96: Same concern as above

package.json
```diff
@@ -111,10 +111,10 @@
...
     "@electron-forge/plugin-webpack": "6.0.0-beta.54",
     "@marshallofsound/webpack-asset-relocator-loader": "0.5.0",
-    "@storybook/addon-actions": "^6.2.9",
-    "@storybook/addon-essentials": "^6.2.9",
-    "@storybook/addon-links": "^6.2.9",
-    "@storybook/react": "^6.2.9",
+    "@storybook/addon-actions": "6.2.9",
+    "@storybook/addon-essentials": "6.2.9",
+    "@storybook/addon-links": "6.2.9",
+    "@storybook/react": "6.2.9",
```
CrazyOptimist, line 117: Same concern as above, it was mistake to merge this commit(adding storybook) into master, but no panic because it does nothing actually.

CrazyOptimist (changes requested):

CrazyOptimist (changes requested):

package.json
ngvtuan: Yes, but the `new_development ` will not merge to the master soon, that's why I have to update this to make sure it's consistent in the master branch.

For now, we haven't used any `tailwind`, I think it will not cause any problem when we merge with the `new_developmnet` branch.

And before we merge the `new_development` branch into the `master` branch. I think we should merge the `master` into `new_development` first and when we do this step. I think we can keep all dependencies in the `new_development` and update/remove un-need dependencies of the `master` branch. By this way, it will avoid confusion and it safe to merge.

package.json
ngvtuan: Please see my comment above

package.json
CrazyOptimist: Yes, remove or upgrade same as `new_development` - take the easier way for you, I agree 💯

package.json
ngvtuan: Yes, I know it is an update to remove `^` to make the package.json consistent.

src/renderer.tsx
```diff
@@ -76,8 +76,6 @@ ipcRenderer.on(
     }: { isPackaged: string; locatePastelConfDir: string },
   ) => {
     if (isPackaged) {
-      console.log = log.log
-      console.error = log.error
```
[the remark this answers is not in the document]
ngvtuan: Why should we use log instead of console.log? 
Because it will use when you want to write something to the log file

is it causing any bug?
Yes, because if we writing all `console` to the log file, it may have sensitive info and we can't control/handle this case. Because we haven't managed the console.log/console.error in the development before which means if someone has used `console` in the code, it will write that data/info to the log file.

In the previous PR I suggested to use console.log instead of log, please advice if I'm missing something important
Yeah it a good suggestion and I updated base on your suggestion, But when I released a new version, I realized if we do like that we can't control the log file and the purpose of the log file is to write useful information, but it must be controlled. It means we need to know what we writing into the log file. So if someone wants to write something to the log file they must know exactly what they want to log.

src/legacy/rpc.ts
```diff
@@ -3,6 +3,7 @@
 import axios from 'axios'
 import _ from 'underscore'
 import hex from 'hex-string'
+import log from 'electron-log'
```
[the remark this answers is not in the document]
ngvtuan: Yes, depending on your config. See here https://www.npmjs.com/package/electron-log at the. `Disable a transport` section

src/renderer.tsx
```diff
@@ -76,8 +76,6 @@ ipcRenderer.on(
...
       log.transports.console.level = false
     }
```
CrazyOptimist, line 80: Could you please add `else` section with 
```
log.transports.file.level = false
```
As we don't need file log for a not packed app, anyway not bad so I'm not blocking

CrazyOptimist (approved):

PR #158, opened 2021-06-18
add radio component and fix wallet screen

sanjanadhikari1:
- [ ] Develop the radio component according to design and added it in the storybook.
- [ ] Use new radio component in the updated Wallet Screen.
image


src/common/components/Radio/Radio.tsx
```diff
@@ -1,20 +1,55 @@
...
 
-function Radio({ children, isChecked, clickHandler }: RadioProps): JSX.Element {
+const Radio = ({
+  children,
+  isChecked,
+  clickHandler,
+  className = 'w-5 h-5',
+  smallCircleClass = 'w-2 h-2',
+}: TRadioProps): JSX.Element => {
+  const [checked, setChecked] = useState<boolean>(isChecked)
```
CrazyOptimist, line 19: We do we need both isChecked from props and local checked?
I guess having only isChecking and clickHandler from props is enough

src/common/components/Radio/Radio.tsx
```diff
@@ -1,20 +1,55 @@
...
+  className?: string
+  smallCircleClass?: string
   children: React.ReactNode
 }
 
-function Radio({ children, isChecked, clickHandler }: RadioProps): JSX.Element {
+const Radio = ({
+  children,
+  isChecked,
+  clickHandler,
```
CrazyOptimist, line 15: Better to have more conventional names, like in standard checkbox or radio inputs: `checked` and `onChange`.
Please update if that sounds good to you

src/common/components/Radio/Radio.tsx
```diff
@@ -1,20 +1,55 @@
...
+    setChecked(isChecked)
+  }, [isChecked])
   return (
-    <Styles.Container className='radioContainer' onClick={clickHandler}>
-      {children}
-      <Styles.Input type='radio' checked={isChecked} onChange={() => null} />
-      <Styles.Span className='radioCheckmark' />
-    </Styles.Container>
+    <div className='flex items-center'>
+      <div
```
CrazyOptimist, line 25: radio input is better than a div: user can navigate to it using tab and toggle it using space

CrazyOptimist (changes requested):

ngvtuan (approved):

src/common/components/Radio/Radio.tsx
sanjanadhikari1: fixed

src/common/components/Radio/Radio.tsx
sanjanadhikari1: fixed

src/common/components/Radio/Radio.tsx
sanjanadhikari1: fixed

CrazyOptimist (approved):

PR #159, opened 2021-06-20
Fix focus issue on expert console after run command

westalert:


src/features/expertConsole/TerminalConsole.tsx
```diff
@@ -76,6 +76,7 @@ function TerminalConsole(props: TConsoleProps): JSX.Element {
   const [isReady, setIsReady] = useState(isRendered)
   const [isInitiatedTerminal, setIsInitiatedTerminal] = useState(false)
   const inputRef = createRef<HTMLInputElement>()
+  const Input_Terminal = document.getElementById('terminalInput')
```
ngvtuan, line 79: Can you change the variable name to `inputTerminal` ? Because I think the name of the variable needs to consistent like other sections in this file.

For example 
- `const inputRef = createRef<HTMLInputElement>()` 
- or https://github.com/pastelnetwork/pastel-electron-wallet/blob/master/src/features/expertConsole/TerminalConsole.tsx Line 117

ngvtuan:
Closed. Because we have created new one here https://github.com/pastelnetwork/pastel-electron-wallet/pull/162 and it is already merged to the master

PR #160, opened 2021-06-20
Feature/searchresults 40285870

ihor-stack:
2.2 Search Result


src/common/components/Header/Header.tsx
```diff
@@ -54,8 +65,19 @@ const SearchBar = () => {
...
         />
       </div>
+
+      <SearchResults
+        open={openSearchResults}
+        onClose={toggleSearchResults}
+        foundNFTs={mockNFTs}
+        foundMembers={mockMembers}
+        onChangeTimeRange={() => {
+          /**/
```
ngvtuan, line 78: Can we remove this line?

src/features/searchResults/mockData.ts
```diff
@@ -0,0 +1,78 @@
+import { TMemberCard } from 'features/members/MemberCard'
+import { TNFTCard } from '../../common/components/NFTCard'
+import mockAvatar from '../../common/assets/images/avatar2-placeholder.png'
+import image from '../../common/assets/images/nft-card-placeholder.png'
```
ngvtuan, line 4: Can we change the variable name to `nftCard`?

ngvtuan (commented):
Issue 1
Please reduce the width of the pop up

Issue 2
1. Reduce the padding of 2 items
2. Can we add 3 rows like the design?

CrazyOptimist (approved):

src/features/searchResults/mockData.ts
ihor-stack: We discussed as "T". Need to change again?

ngvtuan: @ihor-stack I think we just need change the name from `image` to `nftCard` is ok.

src/common/components/Header/Header.tsx
ihor-stack: If I remove comment line, eslint will warn about empty func body.

ngvtuan (approved):

PR #176, opened 2021-06-30
Feature/added change username modal 40754010

sanjanadhikari1:
- [ ] Added ChangeUsername Modal and linked it.

[https://freedcamp.com/view/2956183/tasks/40754010](url)


src/common/components/Buttons/Button.tsx
```diff
@@ -5,6 +5,7 @@ export type TButton = {
   children?: ReactNode | string
   variant?: 'default' | 'secondary' | 'transparent' | 'navigation'
   disabled?: boolean
+  disabledBule?: boolean
```
CrazyOptimist, line 8: I'm afraid that no one will guess what is "disabledBule", and it's important to have understandable props in shared components.

Please remove this prop and pass "bg-opacity-50" and other classes you need via `className` prop

CrazyOptimist (changes requested):

src/common/components/Buttons/Button.tsx
sanjanadhikari1: fixed

CrazyOptimist (approved):

ngvtuan (commented):
ui issue

PR #181, opened 2021-07-05
Feature/import export address labels 40895968

ngvtuan:
##### Commits
- [x] We need to include in your QRcode videos any user defined address labels

##### FreedCamp
- [x] [We need to include in your QRcode videos any user defined address labels](https://freedcamp.com/view/2956183/tasks/40895968)


sanjanadhikari1 (approved):
It looks good for me

goal-shevchenko (approved):
It looks good to me.

PR #193, opened 2021-07-26
Chore/UI fixes jul 26

CrazyOptimist:
1) Wallet startup aspect ratio should be 4:3
2) Change price formatting, currently only on members page: so 1 -> 1 PSL, 1000 -> 1K PSL, 1000000 -> 1M PSL, 1000000000 -> 1B PSL
3) Always show tooltip for slider in nft info modal(only here, not other parts in wallet)
4) Modify "Upload" to "Select" in Add NFT modals, we never upload image before compression.
5) Improve button paddings in Add NFT modals to have paddings as on design. Needed to add classes .btn and .btn-primary because existing Button component is not good enough for styling customization
6) For image size, remove space between digit and unit. "157 MB" -> "157MB"

Note:
Branched from `feature/image-optimization-png-jpg` to avoid conflicts
#190 will be merged first, and then #180 and #193, to reduce dup work on resolving conflicts.


sanjanadhikari1 (commented):
It seems some of the changes overlapped to the PR 190. We need to create Pr for each future you know. Other than that looks good for me!

CrazyOptimist:
> It seems some of the changes overlapped to the PR 190. We need to create Pr for each future you know. Other than that looks good for me!

Yes, the reason is in the PR description.
@sanjanadhikari1

ngvtuan (approved):
Look good to me

sanjanadhikari1 (approved):

PR #200, opened 2021-07-29
Feature/nsfw nft

CrazyOptimist:
##### Done
- [x] Add hentai and porn metrics (from 0 to 100) to nfts data
- [x] Add sliders to user profile for selecting 0-100 value to filter images
- [x] Blur images with higher values if hiding NSFW is turned on in settings, it is on by default

##### Note
No designs for this currently, so please check only logic

##### Screenshot(s)
image


src/common/components/NFTCard/NFTCard.tsx
```diff
@@ -159,20 +163,23 @@ const NFTCard = ({
...
         </div>
       ) : null}
       <Link to={detailUrl} className='cursor-pointer w-full'>
         <div
           className={cn(
-            'relative',
+            'relative overflow-hidden',
             isPortfolio && 'h-220px',
             !isPortfolio && imageHeightClass,
           )}
```
sanjanadhikari1, line 175: I think it's better to use isPortfolio? 'h-220px': imageHeightClass. Can we change this?

sanjanadhikari1 (changes requested):
Others looks good to me

src/common/components/NFTCard/NFTCard.tsx
CrazyOptimist: Fixed, though I don't think it is better and this wasn't changed by this PR

sanjanadhikari1: okay I can see this is not belongs to PR. Anyway thank you!

sanjanadhikari1 (approved):

ngvtuan (approved):

PR #213, opened 2021-08-16
wallet screen updates

ngvtuan:
##### Fixed
- [x] added classes for the Typography component
- [x] added QR code popup when hovering to the QR code icon
- [x] updated Total select amount when user changed the amount in the payment modal
- [x] updated the select total when the user changes the dropdown value
- [x] fixed UI issues for the wallet screen
- [x] created functionality to show the items of the dropdown next to the balance base on the PSL amount
- [x] implement the copy feature in the generate paper wallet popup
- [x] create functionality for Generate a new PSL address button
- [x] fixed UI for the transaction history popup
- [x] updated UI for the note popup
- [x] fixed Warning: Encountered two children with the same key
- [x] fixed UI for the payment popup
- [x] implement functionality to get the address from the address list in the wallet screen and show it to the payment popup
- [x] show amount of PSL in the Confirm PSL ... payment button
- [x] implement functionality to show/hide the empty addresses
- [x] removed eye icon, added hover, active state for copy icon
- [x] added copy functionality
- [x] updated pdf download feature, remove un-need code
- [x] fixed the Selected total is not updated if the user added amount into the dropdown
- [x] implement for show/hide empty addresses functionality
- [x] updated format for the last activity column
- [x] added hover state for table
- [x] added hover/active state for icons
- [x] added tooltip
- [x] updated text for comment modal
- [x] added hover/active state for icons


src/common/components/Inputs/InputExportKey.tsx
```diff
@@ -1,46 +1,41 @@
-import React from 'react'
+import React, { useState } from 'react'
 import SVG from 'react-inlinesvg'
+import { clipboard } from 'electron'
+
 import Input, { TInput } from './Input'
-import EyeIcon from '../../assets/icons/ico-eye.svg'
-import EyeIconHidden from '../../assets/icons/ico-eye-hidden.svg'
 import PasteIcon from '../../assets/icons/ico-paste.svg'
+import CheckIcon from '../../assets/icons/ico-check-green.svg'
```
sanjanadhikari1, line 7: I think this check icon is already in the common/components. Please check!

src/features/wallet/CommentModal.tsx
```diff
@@ -1,51 +1,78 @@
 import React, { useState } from 'react'
 import cx from 'classnames'
+
+import Tooltip from 'common/components/Tooltip'
+import { EliminationIcon } from 'common/components/Icons'
+import Typography, { TypographyVariant } from 'common/components/Typography'
 import styles from './CommentModal.module.css'
-import infoIcon from '../../common/assets/icons/ico-info.svg'
 
 export interface CommentModalProps {
```
sanjanadhikari1, line 9: Can we change this to type?

sanjanadhikari1 (changes requested):

src/features/wallet/CommentModal.tsx
ngvtuan: updated

src/common/components/Inputs/InputExportKey.tsx
ngvtuan: updated

src/index.css
```diff
@@ -176,6 +176,126 @@
     -webkit-background-clip: text;
     -webkit-text-fill-color: transparent;
   }
+
+  .h1_32_40_heavy {
```
CrazyOptimist, line 180: I see can see a different naming here:
1) `.h1_32_40_heavy`
2) `.mini_heavy`
3) `.text_12_16_roman`

And I'd like to suggest another naming instead:

`text-h1-heavy`
`text-h2`
`text-mini-italic`

First part to indicate this is a class for text
Second part is a size name from design. In design h1, h2, mini already means size, so it should be safe to omit 16/24 exact part.
And third part is for font-style property: heavy, italic. I think better to omit `roman` because it means "normal" and the font is already normal by default. Design has "itali**a**c", you renamed it to "italic" I guess for the same reason: in web dev "italic" word is more common.

CrazyOptimist (changes requested):

src/common/components/Modal/TitleModal.tsx
```diff
@@ -21,11 +22,15 @@ const TitleModal = ({
...
+  classNames,
 }: TModalProps): JSX.Element => {
-  const modalClasses = cn({
-    'z-50 relative bg-white rounded-2xl shadow-xSmall w-full max-h-full py-8 overflow-auto mx-auto focus:outline-none': true,
-    [`max-w-${size}`]: size,
-  })
+  const modalClasses = cn(
+    {
+      'z-50 relative bg-white rounded-2xl shadow-xSmall w-full max-h-full pt-10 pb-30px overflow-auto mx-auto focus:outline-none': true,
+      [`max-w-${size}`]: size,
```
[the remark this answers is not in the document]
ngvtuan: updated.

src/common/components/TimeAgo/TimeAgo.tsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react'
+import cn from 'classnames'
+
+export type TTimeAgoProps = {
+  date: number
+  className?: string
+}
+
+export default function TimeAgo({
```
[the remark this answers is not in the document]
ngvtuan: updated

src/index.css
ngvtuan: updated

sanjanadhikari1 (changes requested):

src/features/wallet/PaymentSource.tsx
```diff
@@ -1,42 +1,64 @@
...
+import PastelUtils from 'common/utils/utils'
+import { formatAddress, parseFormattedNumber } from 'common/utils/format'
 import Checkbox from '../../common/components/Checkbox/Checkbox'
-import pasteIcon from '../../common/assets/icons/ico-paste.svg'
-import pencilIcon from '../../common/assets/icons/ico-pencil.svg'
-import passEyeIcon from '../../common/assets/icons/ico-pass-eye.svg'
 import add2Icon from '../../common/assets/icons/ico-add-2.svg'
-import Select from '../../common/components/Select/Select'
-import CommentModal from './CommentModal'
+import SelectAmount, { TOption } from '../../common/components/SelectAmount'
```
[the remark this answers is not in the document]
CrazyOptimist: @sanjanadhikari1 When importing neighbour file better to use relative import: `./file`, but when importing file from the far it's better to use absolute import `common/utils/file`. I don't think that editors are smart enough to do that automatically, and fixing imports manually makes a little sense, so I suggest to ignore it

src/index.css
```diff
@@ -177,125 +177,109 @@
...
     @apply text-h4 leading-6 font-medium;
   }
 
-  .h4_18_24_roman {
+  .text_h4 {
     @apply text-h4 leading-6 font-normal;
   }
 
-  .h5_16_24_heavy {
+  .text_h4_leading_28_heavy {
```
CrazyOptimist, line 208: Could you remove classes with `leading`?
First because I can't see h4 with different line height in design
Second because better to have a standard class for h4, like `text-h4-heavy`, and then you can add other classes: `className='text-h4-heavy leading-[28px]'` in case if in some place of design it is different

src/index.css
```diff
@@ -177,125 +177,109 @@
     -webkit-text-fill-color: transparent;
   }
 
-  .h1_32_40_heavy {
+  .text_h1_heavy {
```
CrazyOptimist, line 180: I know underscore may look more readable. But please let's use hyphen `-` anyway, since all other classes are named in such way.

src/common/components/TimeAgo/TimeAgo.ts
```diff
@@ -1,15 +1,4 @@
...
-export type TTimeAgoProps = {
-  date: number
-  className?: string
-}
-
-export default function TimeAgo({
-  date,
-  className,
-}: TTimeAgoProps): JSX.Element {
+export default function timeAgo(date: number): string {
```
CrazyOptimist, line 1: `common/components` is only for components, as `timeAgo` is an utility function now, move it to `common/utilities/format.ts` - `format` seems to be close by the meaning.

CrazyOptimist (changes requested):

src/features/wallet/PaymentSource.tsx
```diff
@@ -1,42 +1,64 @@
...
+import { TRow } from 'common/components/Table'
+
+import { Trash, Eye, Pencil } from 'common/components/Icons'
 
 export type IDataType = {
```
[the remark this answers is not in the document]
ngvtuan: Updated

src/index.css
ngvtuan: @CrazyOptimist 

No, we can't remove it. 
First --> It has some places that have been used the different line-height. See the attachment ( line-height.jpg) 
line-height and I think it hasn't been defined in the typography yet.
Second --> We can't use `className='text-h4-heavy leading-[28px]'` to overwrite the line-height in the `text-h4-heavy`

src/index.css
ngvtuan: Updated

src/common/components/TimeAgo/TimeAgo.ts
ngvtuan: Updated

sanjanadhikari1 (approved):

src/index.css
CrazyOptimist: We can use `className='text-h4-heavy leading-[28px]'` if move text classes from `utilities` to `components` layer, so other utility classes will have precedence

CrazyOptimist (approved):

src/index.css
ngvtuan: OK I will check and update in the next PR

PR #215, opened 2021-08-17
Add caching for wallet screen

CrazyOptimist:
Added [react-query](https://react-query.tanstack.com/overview) library which helps to perform queries and it keeps a cache

Goal is to speed up loading of wallet screen. The screen depends on balances and addresses.

Balances and pastel price is fetched in interval by `PastelDBThread`, I modified the Thread to update cache of react-query for balances and to update redux store for pastel price when requests succeeded.

Addresses are fetched once app is started by `RendererSetupHooks` component.

On wallet page I used a Spinner component to indicate if data wasn't loaded yet in case if it will take to long to the Thread to load data, but normally I expect user to see balances and addresses immediately once page is opened

Removed the condition to load data in interval only when app is not packed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,6 +104,7 @@
     "react-popper": "2.2.5",
     "react-popper-tooltip": "4.3.0",
     "react-qr-reader": "2.2.1",
+    "react-query": "^3.19.6",
     "react-range": "1.8.8",
     "react-ratings-declarative": "3.4.1",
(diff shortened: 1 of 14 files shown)


sanjanadhikari1 (approved):
It looks good to me.

Dicklesworthstone (approved):

PR #225, opened 2021-09-02
Chore/optimize wallet screen

CrazyOptimist:
Imported sqlite database from file when opening app and write it back when closing

Since this point we won't be able to modify columns easily, because db is persistent now, so applied fixes to db tables:
- renamed create_timestamp columns to createdAt
- changed id to be auto incremented primary key column on all tables, removed manually fetching id before insert

Added `throw` option to rpc which makes it throw error when error happens and returns result when no error.

Add loading transactions to `pastelDbThread`, transactions are loaded in batches.

Wallet screen:
- consume state in all components via react context to avoid 'props drilling'
- load addresses, balances, transactions independently. So addresses can be displayed before others, balances and last activity will be displayed once loaded.
- mount modals only when they are open, because before transaction history modal was loading transactions even in closed state.


sanjanadhikari1 (dismissed):
It works great and looks good to me.

ngvtuan (dismissed):

sanjanadhikari1 (approved):

Dicklesworthstone (approved):

PR #231, opened 2021-09-13
Feature/onboarding integration

CrazyOptimist:
After submitting registration by promo code it will wait for 1 confirmation of pastel id registration ticket and then will open dashboard
Waiting for confirmation means to call rpc in 1 second interval

When app starts, it will check if user has pastel id. User may have multiple ids, and by current implementation it takes first pastel id to use.
If user has no pastel id it will open welcome screen with the button to register
If user has pastel id, it will check if pastel id has confirmations and if no it will open "registration pending" screen and will check for confirmations in the interval
And if user has confirmed pastel id and it was confirmed the app will show login screen.

Login screen now is really checking password to let user in. So username is still ignored, but password is verified.
To not block developers I've added .env.example file. Copy .env.example to .env and write such variables:
```
AUTO_LOGIN_USERNAME=any_string
AUTO_LOGIN_PASSWORD=any_string
```
Currently, while we don't have username integration, you can specify any string in both variables and Onboarding screen will be skipped. There variables have effect only in development mode

Added api code for walletnode userdatas api endpoints - not used for now, this api is for profile info only, such as location, bio, but it is not related to username

Moved onboarding routes to the onboarding feature file, because now Onboarding.tsx controls which screen to show in onboarding.

Added zustand state manager to clearing out the code in onboarding. In other sections I've used React Context for the same purpose, but React Context has drawbacks of re-rendering components on any change, while zustand has selectors, so component gets re-rendered only on specific data update. Redux could be used for this too, but zustand is way simpler and I hope it will help.


package.json
```diff
@@ -61,8 +61,9 @@
     "currency-symbol-map": "5.0.1",
     "d3": "7.0.0",
     "dateformat": "3.0.3",
-    "dayjs": "1.10.4",
+    "dayjs": "1.10.7",
     "detect-port": "1.3.0",
+    "dotenv-webpack": "^7.0.3",
```
sanjanadhikari1, line 66: Can we use the specific version for this?

sanjanadhikari1 (changes requested):

package.json
CrazyOptimist: Thanks for catching!

ngvtuan (commented):
1. Console error `13:11:26.842 › ERROR: error starting server: listen tcp 127.0.0.1:8080: bind: Only one usage of each socket address (protocol/network address/port) is normally permitted.`
2. When I removed all pastel folders in my local and run the project from scratch. I getting this error. 

```
13:30:44.452 > Error: Cannot find the Pastel network parameters in the following directory:
13:30:44.458 > "C:\Users\[username]\AppData\Roaming\PastelParams"
13:30:44.459 > Please run 'pastel-fetch-params' or './pcutil/fetch-params.sh' and then restart.
13:30:45.821 > error: couldn't connect to server: EOF reached (code 1)
13:30:45.823 > (make sure server is running and you are connecting to the correct RPC port)
13:30:57.917 > error: couldn't connect to server: EOF reached (code 1)
13:30:57.918 > (make sure server is running and you are connecting to the correct RPC port)
13:31:10.025 > error: couldn't connect to server: EOF reached (code 1)
13:31:10.026 > (make sure server is running and you are connecting to the correct RPC port)
13:31:22.163 > error: couldn't connect to server: EOF reached (code 1)
13:31:22.164 > (make sure server is running and you are connecting to the correct RPC port)
13:31:34.273 > error: couldn't connect to server: EOF reached (code 1)
13:31:34.274 > (make sure server is running and you are connecting to the correct RPC port)
13:31:46.391 > error: couldn't connect to server: EOF reached (code 1)
13:31:46.392 > (make sure server is running and you are connecting to the correct RPC port)
13:31:58.491 > error: couldn't connect to server: EOF reached (code 1)
13:31:58.492 > (make sure server is running and you are connecting to the correct RPC port)
13:32:10.592 > error: couldn't connect to server: EOF reached (code 1)
13:32:10.593 > (make sure server is running and you are connecting to the correct RPC port)
13:32:22.689 > error: couldn't connect to server: EOF reached (code 1)
13:32:22.690 > (make sure server is running and you are connecting to the correct RPC port)
13:32:34.797 > error: couldn't connect to server: EOF reached (code 1)
13:32:34.798 > (make sure server is running and you are connecting to the correct RPC port)
13:32:44.814 > [Sep 15 13:32:44.814] ERROR Start walletnode: pasteld didn't start error=pasteld was not started
13:32:44.816 > [Sep 15 13:32:44.814] ERROR Start walletnode: pasteld failed to start error=pasteld was not started
13:32:44.823 > [Sep 15 13:32:44.823] FATAL pasteld was not started
13:32:44.837 > Error: ERROR: pasteld was not started
```
To fix this issue I have copied all pastel folders in the `C:\Users\[username]`  to `C:\Users\[username]\AppData\Roaming` but I think we need to update the code if we moved the pastel folders location.

3. I'm getting this error after fixed for 2. 
```
13:39:13.579 › api/pastel-rpc server error. Response: {"result":null,"error":{"code":-1,"message":"boost::filesystem::directory_iterator::construct: The system cannot find the path specified: \"C:\\Users\\[username]\\Pastel\\pastelkeys\""},"id":"pastelid"}. Status code: 500
```
To fix this issue, I have created `pastelkeys` folder in `C:\\Users\\[username]\\Pastel\\` but I think we need to create the pastelkeys folder when we download the pastel.

4. When I pasted my key in the `Pastel Promo Code` field and then click on the Apply button. The status of that button is not changing. I think we must disable the apply button and change the text to `Applying`.

ngvtuan (approved):

Dicklesworthstone (approved):

